PR #2, opened 2020-03-01
MapBox and Threading Update

ADSteele916:
Switched from GoogleMaps to MapBox map tiles. Updated map point and tile data representation to better work with MapBox. Added threading to prevent freezing when redrawing the map.


MapBox.py
```diff
@@ -0,0 +1,226 @@
...
+    maps = None
+
+
+class MapPoint:
+
+    def __init__(self, latitude, longitude):
+        self.latitude = latitude
+        self.longitude = longitude
+        self.x = self.getX()
+        self.y = self.getY()
```
SilentSys: Is having self.x and self.getX() redundant?

MapBox.py
```diff
@@ -0,0 +1,226 @@
...
+
+class MapTile:
+
+    def __init__(self, x, y, s):
+        self.x = x
+        self.y = y
+        self.s = s
+
+    def getName(self):
+        return f"{self.x}___{self.y}"
```
SilentSys: I'm kinda nitpicking, but could we do with just one "_"?
```suggestion
        return f"{self.x}_{self.y}"
```

main.py
```diff
@@ -92,19 +106,26 @@ def receiveData(self, bytes): #TODO: ARE WE SURE THIS IS THREAD SAFE? USE QUEUE
...
+        #     self.plotMap(latitude, longitude)  # Uncomment to make map recenter
+        #     self.lastMapUpdate = time.time()
+        #     self.lastLatitude = latitude
+        #     self.lastLongitude = longitude
+        # # self.plotMap(latitude, longitude) #Uncomment to make map recenter
+        #
+        # newtime = time.time()
+        # if newtime - self.lastgps >= 3:
+        #     self.updateMark(latitude, longitude)
+        # self.lastgps = newtime
```
SilentSys: Is any of this still useful?
```suggestion
```

main.py
```diff
@@ -114,25 +135,40 @@ def sendCommand(self, command):
...
     def plotMap(self, latitude, longitude):
-        p = GoogleMaps.MapPoint(longitude, latitude)
-        if longitude is None or latitude is None or (p.getTileX(self.zoom) == self.xtile
-                                                               and p.getTileY(self.zoom) == self.ytile):
-            return
+        p = MapBox.MapPoint(latitude, longitude)
+
+        with self.lock:
+            if longitude is None or latitude is None or p.x == self.x and p.y == self.y:
+                return
```
SilentSys: Is this redundant if we already have self.longitude and self.latitude?

PR #3, opened 2020-03-23
Autosave

ADungate:
Solved conflicts between autosave and master, should be good to go.


main.py
```diff
@@ -19,7 +19,12 @@
 import MapBox
 import RocketData
 from RocketData import RocketData as RD
-import mplwidget  # DO NOT REMOVE pyinstller needs this
+
+import _thread
+import asyncio
+import mplwidget #DO NOT REMOVE pyinstller needs this
+
+import threading
```
SilentSys, line 27:
```suggestion
import mplwidget #DO NOT REMOVE pyinstller needs this
```
I don't think you need any of this anymore

main.py
```diff
@@ -38,6 +43,7 @@
 
 TILES = 14
 
+lock = asyncio.Lock()
```
SilentSys, line 46:
```suggestion
```

Doesn't seem used

main.py
```diff
@@ -65,6 +71,9 @@ def __init__(self, connection):
 
         self.sendButton.clicked.connect(self.sendButtonPressed)
         self.commandEdit.returnPressed.connect(self.sendButtonPressed)
+
+        # TODO: Create auto saving thread
+        #thread.start_new_thread(self.actionSave.triggered.connect, self.data.save)
```
SilentSys, line 76:
```suggestion
```

TODO = done

RocketData.py
```diff
@@ -53,66 +54,82 @@
...
 }
-
 class RocketData:
     def __init__(self):
         self.timeset = {}
         self.lasttime = 0
         self.highest_altitude = 0
+        self.sessionName = str(int(time.time()))
+        t = threading.Thread(target=self.timer)
+        t.start()
```
SilentSys, line 64:
```suggestion
        self.autosaveThread = threading.Thread(target=self.timer)
        self.autosaveThread.start()
```

It is usually a good idea to keep a reference to all threads.

PR #4, opened 2020-03-30
Radio Protocol Parsing implementation

DSep:
Implementation of structure for new Radio Protocol parsing. Current implementation includes parsing of BulkSensor subpackets only, but provide structure to easily add others.

Also added a few more things that should help future work.

(Pull request is created as a way of getting active feedback on branch, but I do recognize code still needs cleaning up and also work on functioning side by side with the mapping)


DSep:
I added a lot of type hints as well, so let me know what you think about that.

RadioController.py
```diff
@@ -0,0 +1,112 @@
...
+for i in SubpacketIDs.get_list_of_sensor_IDs():
+    PACKET_ID_TO_CONST_LENGTH[i] = 5
+
+
+# Check if packet of given type has constant length
+def isPacketLengthConst(subpacket_id):
+    return subpacket_id in PACKET_ID_TO_CONST_LENGTH.keys()
+
+
+# Return dict of subpacket id, length of subpacket in byte_list, subpacket typename, data_unit
```
SilentSys, line 24: Wait, its returning 4 things?

RadioController.py
```diff
@@ -0,0 +1,112 @@
...
+        data_unit = byte_list[1:length]
+        del byte_list[0:1]
+        length = length - 1  # TODO Fix this bad code somehow
+    else:
+        length = int.from_bytes(byte_list[1], "big")
+        data_unit = byte_list[2:length]
+        del byte_list[0:2]
+        length = length - 2  # TODO Fix this bad code somehow
+    parsed_subpacket: Dict[str, int] = {'id': subpacket_id, 'length': length}
+    return parsed_subpacket
```
SilentSys, line 43: It looks like this could just be:
```suggestion
    return subpacket_id, length
```

and then calling it would just be:
`subpacket_id, length = extract_subpacket(byte_list)`

Seems simpler and python always does stuff like this in its libraries I think. Although what I like about the dict is that you dont have to worry about ordering (but you do have to worry about strings ig).

Either way works!

RadioController.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+# general data parser interface
+def parse_data(type_id, byte_list, length):
+    return packetTypeToParser[type_id](byte_list, length)
```
SilentSys, line 48: Love this

SerialThread.py
```diff
@@ -75,45 +81,72 @@ def trySendMessage(self):
...
+                    continue
+                except Exception as e:
+                    print(e)
+                    del byteList[0:1]
+                    continue
+
+                chunk_length = data_and_info['length']
+                data = byteList[0:chunk_length]  # convert all of the data in the chunk to byte
+
+                parsed_subpacket_data = RadioController.parse_data( data_and_info['id'], data, chunk_length )
```
SilentSys, line 125: Would it make more sense if we put this inside of extract_subpacket (or inside another method containing them both) so that way we're not going "in and out" of RadioController twice?

Or would that make things worse? Not sure.

RocketData.py
```diff
@@ -109,35 +136,43 @@ def save(self):
...
+#
+#             if datatype == "int":
+#                 return toint(data)
+#             elif datatype == "state":
+#                 return tostate(data)
+#
+#         return fourtofloat(data)
+#
+#     except:
+#         return -1
```
SilentSys, line 178: Bye bye old code
```suggestion
```

DSep:
Note: Outcome of discussion over this PR is that all subpackets should come with a timestamp from the rocket.

PR #5, opened 2020-05-06
Pipelined reading/writing to IConnection

SilentSys:


main.py
```diff
@@ -76,20 +78,26 @@ def __init__(self, connection):
...
+        self.SendThread.sig_print.connect(self.printToConsole)
+        self.SendThread.start()
 
         # Mapping thread init and start
         thread = threading.Thread(target=self.threadLoop, daemon=True)
         thread.start()
 
     # Possible doc: receives from the serial thread loop of subpackets
-    def receiveData(self, dataBundle):  # TODO: ARE WE SURE THIS IS THREAD SAFE? USE QUEUE OR PUT IN SERIAL THREAD
+    def receiveData(self, dataBundle):
```
DSep, line 100: For this part of data flow: What happens when the incoming data is dequeued through the new system and this is called? Will this (and the line below) properly queue up adding new data to RocketData regardless of the availability of the RocketData mutex lock?

ReadThread.py
```diff
@@ -0,0 +1,76 @@
...
+
+from PyQt5 import QtCore
+from PyQt5.QtCore import pyqtSignal
+import queue
+
+import RadioController
+
+class ReadThread(QtCore.QThread):
+    # sig_received = pyqtSignal(list)  # TODO Review this change with Andrei
+    sig_received = pyqtSignal(object)  # TODO is it safe to generalize
```
DSep, line 11: A bit of an artifact but useful to ask now especially with this change: Is it reasonable to have this generalization? It was necessary to change it for my earlier changes due to the way that data is now being sent back as dictionary but not a list anymore.

main.py
SilentSys: It will block the ReadThread on the next line until the mutex is free. But that's okay because its what that thread is for. We get a large advantage by implementing this thread and queue because we are essentially pipelining /parallelizing the CPU bound tasks with the IO bound tasks, so this allows each to work independently of each other allowing us to get ALL the data we can from the XBee. We could add multiple queues and processing threads at any point along the flow, but that would be parallelizing CPU bound tasks with other CPU bound tasks which is subject to diminishing returns and could actually make processing slower if we reach a CPU or IO throughput limit (especially if the threads are fighting for CPU time, but that's not the case for us, we are limited by IO/Xbee).

It would be interesting to monitor the size of the queues. If they are growing over time, then we know for sure that we need to improve the performance/parallelism of our data processing. But with how fast CPUs are these days and how slow the radio communication is, I'm pretty certain that the queue sizes would remain around 0-1 indicating that the IO/radio is the bottleneck.

It would be cool to maybe add a warning message if the queue is somehow getting too big.

SilentSys:
Sweet! I'll comment some stuff and update the PR soon.

ReadThread.py
SilentSys: Just looked into it a bit. That should be fine.

ReadThread.py
```diff
@@ -0,0 +1,76 @@
...
+        self.errored = False
+
+    def _newData(self, data):
+        self.dataQueue.put_nowait(data)
+
+    # Function that is run by the thread when it is not blocked
+    def run(self):
+        self.running = True  # TODO Review purpose of this
+        # Loop that attempts a message send and gets+parses data repeatedly
+        while self.running:
```
DSep, line 30: Another artifact but thought it would be relevant to discuss:
What is the point of having this as a variable?
And if it is a field, why declare it here? Is that good practice?

SilentSys: Usually that's how you'd terminate a thread. There is supposed to be a stop() function that sets running to false. But we don't ever do that so that code is obsolete.

Yeah it should have been "declared" in init imho. But I'll just remove it entirely.

DSep:
Thank you! I think those function descriptions will make reading much easier and faster, and the discussion in this PR can be useful for future reference!

PR #6, opened 2020-05-11
Sim connection

SilentSys:
We can now connect the GS to the FW! Although, most of the radio communication protocol still hasn't been implemented, so functionality is limited.


main.py
```diff
@@ -44,55 +37,66 @@ class MainApp(QtWidgets.QMainWindow, Ui_MainWindow):
...
+
+    def closeEvent(self, event):
+        self.connection.shutDown()
+        print("Saving...")
+        self.data.save(os.path.join(LOCAL, "finalsave_" + str(int(time.time())) + ".csv"))
+        print("Saved!")
 
-    # Possible doc: receives from the serial thread loop of subpackets
-    def receiveData(self, dataBundle):  # TODO: ARE WE SURE THIS IS THREAD SAFE? USE QUEUE OR PUT IN SERIAL THREAD
-        self.data.addBundle(dataBundle)
```
SilentSys, line 94: @ADSteele916, main.py is no longer adds data to RocketData. This means that main.py isn't responsible for anything other than the UI now (aside from some set-up at init). Hopefully this can help you with your tests as running without a UI should be much easier now.

ADSteele916: That will be welcome. 👍

main.py
```diff
@@ -212,23 +235,17 @@ def threadLoop(self):
...
+                lastLatitude = lastLat
+                lastLongitude = lastLon
+                lastMapUpdate = lmu
+                lgps = lastgps
 
-                if time.time() - lgps >= 1:
-                    self.updateMark(self.latitude, self.longitude)
+                if time.time() - lgps >= 1:  # TODO sort out what this code does and if it can handle change to way it uses data
+                    with self.lock:
+                        self.updateMark(self.latitude, self.longitude)
```
ADSteele916, line 245: What exactly needs to be sorted out here?

SilentSys: I believe that was a personal note/reminder from @DSep's commit. I kind of based my changes off of his working branch before he finished his work. I was trying to minimize how many conflicts I'd create in his branch because his work is closely related.

SilentSys: You don't need to sort anything out dw.

ADSteele916: Sounds good.

RadioController.py
```diff
@@ -38,92 +37,112 @@ def isPacketLengthConst(subpacket_id):
...
+        c = struct.unpack('>f' if self.bigEndianFloats else '<f', b)
+        return c[0]
+
+    def fourtoint(self, bytes):
+        assert len(bytes) == 4
+        data = bytes
+        # data = data[::-1]#flips bytes
+        b = struct.pack('4B', *data)
+        # big endian
+        c = struct.unpack('>I' if self.bigEndianInts else '<I', b)
```
DSep, line 147: Makes more sense!

main.py
DSep: Yeah, it was a note to self to sort out how data was handled here.

main.py
```diff
@@ -44,55 +37,66 @@ class MainApp(QtWidgets.QMainWindow, Ui_MainWindow):
...
-        self.zoom = 20
-        self.radius = 0.1
 
+        # lock used for lat & lon fields # TODO Remove once RocketData refactored to have lock
         self.lock = threading.Lock()
-        self.lastgps = time.time()
-        self.lastMapUpdate = time.time()
-        self.latitude = None
+
+        self.latitude = None    # NOTE copies of what is already in RocketData TODO refactor to get from RocketData
```
DSep, line 46: In the rest of my branch I was looking to change this to no longer being a part of main.py at all, and just retrieved from RocketData.

Should I avoid this change now, and just leave it as it is? I wasn't sure if it would be wise for the threadLoop to wait on the UI updater receiveData for the new lat and Lon. (And that is the only place using those it seems)

RadioController.py
SilentSys: Oops, I should probably get rid of those old comments tho.

main.py
SilentSys: I DEFINITELY still think there's a lot of refactoring potential. So 100% continue with your work if you're still down. Sorry for changing so much in the area though. It think you could continue with your branch by pulling the changes from master. Or you could start a new branch, should have the same effect in the end.

> In the rest of my branch I was looking to change this to no longer being a part of main.py at all, and just retrieved from RocketData.

Yeah that sounds great. Maybe then it would be easier to move the map thread into its own class too. That would really clean up main.py.

> I wasn't sure if it would be wise for the threadLoop to wait on the UI updater receiveData for the new lat and Lon. (And that is the only place using those it seems)

Having threads wait when there's no work to do is good. So it sounds fine, but maybe I'm missing something. We can talk more about this.

DSep: Cool, I'm happy to pull from master!
>> In the rest of my branch I was looking to change this to no longer being a part of main.py at all, and just retrieved from RocketData.
> Yeah that sounds great. Maybe then it would be easier to move the map thread into its own class too. That would really clean up main.py.
Alright, I can attempt that in my branch!

>> I wasn't sure if it would be wise for the threadLoop to wait on the UI updater receiveData for the new lat and Lon. (And that is the only place using those it seems)
> Having threads wait when there's no work to do is good. So it sounds fine, but maybe I'm missing something. We can talk more about this.
Good point! We should definitely cover this in discussion, but I think conceptually it's totally fine.

PR #7, opened 2020-05-26
Sepand/refactoring data flow

DSep:
- Removed loop and corresponding plotting and marker code to separate function + thread (MappingThread). Setup receiveMap UI updater in main similar to receiveData.
- Setup MapData file inspired by RocketData, allowing a proper, mutex-locked data source.
- Setup callback system for notifying users of RocketData of changes, and used it with a Condition Variable wait+notify within MappingThread.
- Added many TODOs for things to look over, or for future refactoring.
- Bug fixes for the above changes.

### DIFF
--- a/MapData.py
+++ b/MapData.py
@@ -1,41 +1,31 @@
-import os
 import threading
-from enum import Enum
 
-import matplotlib.pyplot as plt
-from scipy.misc import imresize
 
-from detail import LOCAL
+ZOOM = 'zoom'
+RADIUS = 'radius'
+IMAGE = 'image'
+MARK = 'mark'
+LOCATION = 'location'
 
-class MapDataFieldNamesEnum(Enum):
-    ZOOM = 'zoom'
-    RADIUS = 'radius'
-    IMAGE = 'image'
-    MARKER_PATH = 'marker path'
-    MARKER = 'marker'
-    ANNO_BOX = 'annotation box'
-    LOCATION = 'location'
 
-class MapData:
+class MapDataClass:  # TODO Is there a better way of differentiating from the filename? Its causing trouble when trying to use both
     def __init__(self):
         self.lock = threading.Lock()
 
         # Map UI attributes
-        self.mapData = {}
-        self.mapData[MapDataFieldNamesEnum.ZOOM.value] = 20
-        self.mapData[MapDataFieldNamesEnum.RADIUS.value] = 0.1
-        self.mapData[MapDataFieldNamesEnum.IMAGE.value] = None
-        self.mapData[MapDataFieldNamesEnum.MARKER_PATH.value] = os.path.join(LOCAL, "marker.png")
-        # TODO: imresize removed in latest scipy since it's a duplicate from "Pillow". Update and replace.
-        self.mapData[MapDataFieldNamesEnum.MARKER.value] = imresize(plt.imread(self.mapData[MapDataFieldNamesEnum.MARKER_PATH.value]), (12, 12))
-        self.mapData[MapDataFieldNamesEnum.ANNO_BOX.value] = None
-        self.mapData[MapDataFieldNamesEnum.LOCATION.value] = None
+        self.__setattr__(ZOOM, 20)
+        self.__setattr__(RADIUS, 0.1)
+        self.__setattr__(IMAGE, None)
+        self.__setattr__(MARK, None)
+        self.__setattr__(LOCATION, None)
 
-    def getMap(self):
+    # Get the value of the item from the map by Id
+    def getMapValue(self, valueId):
         with self.lock:
-            return self.mapData
+            return self.__getattribute__(valueId)
 
+    # set the value of the item from the map by Id
     # TODO review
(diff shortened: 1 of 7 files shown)


MapData.py
```diff
@@ -0,0 +1,41 @@
...
+        # Map UI attributes
+        self.mapData = {}
+        self.mapData[MapDataFieldNamesEnum.ZOOM.value] = 20
+        self.mapData[MapDataFieldNamesEnum.RADIUS.value] = 0.1
+        self.mapData[MapDataFieldNamesEnum.IMAGE.value] = None
+        self.mapData[MapDataFieldNamesEnum.MARKER_PATH.value] = os.path.join(LOCAL, "marker.png")
+        # TODO: imresize removed in latest scipy since it's a duplicate from "Pillow". Update and replace.
+        self.mapData[MapDataFieldNamesEnum.MARKER.value] = imresize(plt.imread(self.mapData[MapDataFieldNamesEnum.MARKER_PATH.value]), (12, 12))
+        self.mapData[MapDataFieldNamesEnum.ANNO_BOX.value] = None
+        self.mapData[MapDataFieldNamesEnum.LOCATION.value] = None
```
SilentSys, line 32: At first glance, this looks fine, just another map. But after some consideration, it seems like this map is really just trying to be a POD class/struct. So I think to reduce the overhead of looking up in a map and to keep the code simpler, we should just turn it into a POD class.

But then the question is, do we really need ANOTHER class? MapData was already the map data class, so it would seem like we could just make all of these values fields of MapData. However I'm guessing that maybe the original intent with using the map in the first place was to do a sort of "time capture" so that when we plot the map we ensure that all the info came from the same time? But thats kinda what mutexes are for so I think I have a solution for this. Lets arrange to discus this more.

MapData.py
```diff
@@ -0,0 +1,41 @@
...
+    LOCATION = 'location'
+
+class MapData:
+    def __init__(self):
+        self.lock = threading.Lock()
+
+        # Map UI attributes
+        self.mapData = {}
+        self.mapData[MapDataFieldNamesEnum.ZOOM.value] = 20
+        self.mapData[MapDataFieldNamesEnum.RADIUS.value] = 0.1
```
SilentSys, line 26: These are just constants which are used only in MappingThread I believe. I think it would be better to just have them as constants in that file.

MapData.py
```diff
@@ -0,0 +1,41 @@
...
+        self.mapData[MapDataFieldNamesEnum.IMAGE.value] = None
+        self.mapData[MapDataFieldNamesEnum.MARKER_PATH.value] = os.path.join(LOCAL, "marker.png")
```
SilentSys, line 28: This is only used in one place (one line below). I really dont think it deserves a place in a dictionary or even a field in a class. Maybe just a constant somewhere.

MapData.py
```diff
@@ -0,0 +1,41 @@
...
+        # TODO: imresize removed in latest scipy since it's a duplicate from "Pillow". Update and replace.
+        self.mapData[MapDataFieldNamesEnum.MARKER.value] = imresize(plt.imread(self.mapData[MapDataFieldNamesEnum.MARKER_PATH.value]), (12, 12))
```
SilentSys, line 30: This is also just a constant used only in main I think. Could just make it a constant there.

MapData.py
```diff
@@ -0,0 +1,41 @@
...
+        self.mapData[MapDataFieldNamesEnum.ANNO_BOX.value] = None
```
SilentSys, line 31: Annotation box is a UI element, it shouldnt really be passed around as data. It would be nice to keep it only in main.

Not sure if we'd wanna do the math to create the annotation box in main, or just pass the coordinates to it.

MapData.py
DSep: After discussion today it seems like we have still retained the class with a lot of changes but we can still talk about this further.

MapData.py
```diff
@@ -1,41 +1,31 @@
...
+        self.__setattr__(IMAGE, None)
+        self.__setattr__(MARK, None)
+        self.__setattr__(LOCATION, None)
 
-    def getMap(self):
+    # Get the value of the item from the map by Id
+    def getMapValue(self, valueId):
         with self.lock:
-            return self.mapData
+            return self.__getattribute__(valueId)
```
SilentSys, line 25: This is actually really cool 👍

MapData.py
```diff
@@ -1,41 +1,31 @@
...
 
-import matplotlib.pyplot as plt
-from scipy.misc import imresize
 
-from detail import LOCAL
+ZOOM = 'zoom'
+RADIUS = 'radius'
+IMAGE = 'image'
+MARK = 'mark'
+LOCATION = 'location'
```
SilentSys, line 8:
```suggestion
ZOOM = '_zoom'
RADIUS = '_radius'
IMAGE = '_image'
MARK = '_mark'
LOCATION = '_location'
```
Maybe just to signify that these are private. I dont think ppls IDEs would even pick up on them tho.

PR #10, opened 2020-07-13
Fixed app not closing when using SIM

SilentSys:
We already used daemon threads everywhere else, idk why we didn't do that here too.

### DIFF
--- a/XBeeModuleSim.py
+++ b/XBeeModuleSim.py
@@ -70,9 +70,6 @@ class XBeeModuleSim:
 
         self._rocket_rx_thread.start()
 
-    def shutdown(self):
-        pass
-
     def _unpack(self, q):
         """
         Helper generator that unpacks the iterables in a given queue.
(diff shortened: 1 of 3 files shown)


XBeeModuleSim.py
```diff
@@ -65,45 +65,22 @@ def nop_callback(data):
...
-
         # Queues are IO bound.
-        self._rocket_rx_thread = Thread(
-            target=self._run_rocket_rx, name="xbee_sim_rocket_rx"
-        )
+        self._rocket_rx_thread = Thread(target=self._run_rocket_rx, name="xbee_sim_rocket_rx", daemon=True)
 
         self._rocket_rx_thread.start()
 
     def shutdown(self):
```
jesseli2002, line 73: If you're going to make this a no-op you may as well just delete it entirely and remove any references to it (one in SimConnection in shutDown, one in `teardown_method` in the test file.

jesseli2002: Other than that LGTM.

PR #11, opened 2020-07-14
Moved CPU bound map processing into separate process

SilentSys:
And with this, I think I got the last of the freezing / stuttering out. It's taken a while to get here. Please let me know if you guys still experience anything like it in the future. I like to test by just continuously resizing the window and seeing if it freezes when the map updates or spamming a button and seeing if it stops "clicking" and printing during the map update. 

In #8 I said it "hardly ever stutters", now I want to say it NEVER stutters. 🤞

### DIFF
--- a/MapData.py
+++ b/MapData.py
@@ -5,7 +5,6 @@ ZOOM = '_zoom'
 RADIUS = '_radius'
 IMAGE = '_image'
 MARK = '_mark'
-LOCATION = '_location'
 
 
 class MapDataClass:  # TODO Is there a better way of differentiating from the filename? Its causing trouble when trying to use both
@@ -17,7 +16,6 @@ class MapDataClass:  # TODO Is there a better way of differentiating from the fi
         self.__setattr__(RADIUS, 0.1)
         self.__setattr__(IMAGE, None)
         self.__setattr__(MARK, None)
-        self.__setattr__(LOCATION, None)
 
     # Get the value of the item from the map by Id
     def getMapValue(self, valueId):
--- a/MappingThread.py
+++ b/MappingThread.py
@@ -13,6 +13,8 @@ from scipy.misc import imresize
 import MapData
 from SubpacketIDs import SubpacketEnum
 
+# Scaling is linear so a scale factor of 1 means no scaling (aka 1*x=x)
+SCALE_FACTOR_NO_SCALE = 1
 
 # Mapping work and processing that gets put into MapData, repeatedly as RocketData is updated.
 # Signals the main thread to fetch UI elements in MapDataSimilar to ReadData
@@ -158,9 +160,9 @@ def processMap(requestQueue, resultQueue):
                 # Scale "to fit", maintains map aspect ratio even if it differs from the desired dimensions aspect ratio
                 scaleFactor = min(desiredSize[0] / largeMapImage.shape[0], desiredSize[1] / largeMapImage.shape[1])
             else:
-                scaleFactor = 1
+                scaleFactor = SCALE_FACTOR_NO_SCALE
 
-            scaleFactor = min(scaleFactor, 1)  # You shall not scale the map larger. Waste of memory.
+            scaleFactor = min(scaleFactor, SCALE_FACTOR_NO_SCALE)  # You shall not scale the map larger. Waste of memory.
 
             # Downsizing the map here to the ideal size for the plot reduces the amount of work required in the main
             # thread and thus reduces stuttering
--- a/MappingThread.py
+++ b/MappingThread.py
@@ -168,5 +168,6 @@ def processMap(requestQueue, resultQueue):
             int(scaleFactor * largeMapImage.shape[0]), int(scaleFactor * largeMapImage.shape[1])))
 
             resultQueue.put((resizedMapImage, location.xMin, location.xMax, location.yMin, location.yMax))
-        except:
+        except Exception as ex:
+            print("ERROR: Exception in processMap process: %s" % ex)
             resultQueue.put(None)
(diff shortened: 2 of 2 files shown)


MapData.py
```diff
@@ -17,7 +16,6 @@ def __init__(self):
         self.__setattr__(RADIUS, 0.1)
         self.__setattr__(IMAGE, None)
         self.__setattr__(MARK, None)
-        self.__setattr__(LOCATION, None)
```
DSep, line 20: Becoming a local var in MapThread then? Was that part of the cause?

MappingThread.py
```diff
@@ -62,39 +76,32 @@ def plotMap(self, latitude, longitude):
...
-            # Scale "to fit", maintains map aspect ratio even if it differs from the desired dimensions aspect ratio
-            scaleFactor = min(desiredSize[0] / largeMapImage.shape[0], desiredSize[1] / largeMapImage.shape[1])
-        else:
-            scaleFactor = 1
+        self.requestQueue.put_nowait((p1, p2, zoom, desiredSize))
 
-        scaleFactor = min(scaleFactor, 1) # You shall not scale the map larger. Waste of memory.
+        # NOTE: Passing numpy arrays through any sort of IPC other than plane old shared memory will result in the data
+        # being "pickled" and un-"pickled". This uses more processing time than just using shared memory. Since we're
+        # not worried about sync safety here, shared memory might be faster... but also harder to implement.
```
DSep, line 85: Is this something to address in the future then?

MappingThread.py
```diff
@@ -117,19 +124,50 @@ def run(self):
...
 
-                self.plotMap(latitude, longitude)
-
-                # notify UI that new data is available to be displayed
-                self.sig_received.emit()
+                if self.plotMap(latitude, longitude):
+                    # notify UI that new data is available to be displayed
+                    self.sig_received.emit()
+                else:
+                    continue
```
DSep, line 136: Makes sense, cool

MappingThread.py
```diff
@@ -117,19 +124,50 @@ def run(self):
...
+            scaleFactor = min(scaleFactor, 1)  # You shall not scale the map larger. Waste of memory.
+
+            # Downsizing the map here to the ideal size for the plot reduces the amount of work required in the main
+            # thread and thus reduces stuttering
+            resizedMapImage = imresize(largeMapImage, (
+            int(scaleFactor * largeMapImage.shape[0]), int(scaleFactor * largeMapImage.shape[1])))
+
+            resultQueue.put((resizedMapImage, location.xMin, location.xMax, location.yMin, location.yMax))
+        except Exception as ex:
+            print("ERROR: Exception in processMap process: %s" % ex)
```
DSep, line 172: How 'C' of you 😄

MappingThread.py
```diff
@@ -117,19 +124,50 @@ def run(self):
...
+
+            largeMapImage = location.genStichedMap()
+
+            if desiredSize is not None:
+                # Scale "to fit", maintains map aspect ratio even if it differs from the desired dimensions aspect ratio
+                scaleFactor = min(desiredSize[0] / largeMapImage.shape[0], desiredSize[1] / largeMapImage.shape[1])
+            else:
+                scaleFactor = 1
+
+            scaleFactor = min(scaleFactor, 1)  # You shall not scale the map larger. Waste of memory.
```
DSep, line 163: Might want to document this somewhere like confluence for future modifications? Or make '1' not a magic number -> aka a constant

MapData.py
SilentSys: I actually removed it because there aren't any usages. I can't remember where it was used before, but we much have changed something in the meantime and now its just not used.

MappingThread.py
SilentSys: It's not an issue as it is right now. The numpy array is already scaled to the right size so its not that big. Also shared memory is a fixed size and this numpy array is a variable size so it would add a lot of complications to make that work. I'm not sure it's worth it.

If we can determine that we are still having performance issues and it's because of this, then we can change it for sure.

MappingThread.py
SilentSys: Hm, yeah I think I'll turn it into a constant and add a comment. Good point.

It is linear scaling though so the number 1 isn't _that_ magic.

MappingThread.py
```diff
@@ -117,19 +126,50 @@ def run(self):
...
 
             except Exception as ex:
                 print("Error in map thread loop: %s" % ex)
+
+
+# To be run in a new process as the stitching and resizing is a CPU bound task
+def processMap(requestQueue, resultQueue):
+    while True:
+        try:
+            (p1, p2, zoom, desiredSize) = requestQueue.get()
```
jesseli2002, line 152: Pretty sure parentheses not needed

MappingThread.py
```diff
@@ -62,39 +78,32 @@ def plotMap(self, latitude, longitude):
...
+        # not worried about sync safety here, shared memory might be faster... but also harder to implement.
+        result = self.resultQueue.get()
 
-        # Downsizing the map here to the ideal size for the plot reduces the amount of work required in the main
-        # thread and thus reduces stuttering
-        resizedMapImage = imresize(largeMapImage, (int(scaleFactor * largeMapImage.shape[0]), int(scaleFactor * largeMapImage.shape[1])))
+        if not result:
+            return False
+
+        (resizedMapImage, xMin, xMax, yMin, yMax) = result
```
jesseli2002, line 93: I don't think the parentheses are strictly necessary

MappingThread.py
DSep: That's a fair point. Could go either way then

DSep:
Looks like there needs to be conflict resolution first, before merging.

PR #12, opened 2020-07-19
Add docstring stubs

ADSteele916:
Added docstring stubs so that we can have a standardized documentation style throughout our codebase. Also ran isort to standardize the ordering of our import statements.


DSep:
🎉

PR #14, opened 2020-07-20
Handshake

jesseli2002:
Do a handshake between ground station and firmware in SIM, to allow firmware attaching. By default no opportunity is actually given, but you can uncomment an input line to pause the program while you attach a debugger to the FW.

### DIFF
--- a/SimConnection.py
+++ b/SimConnection.py
@@ -19,13 +19,6 @@ LOG_HISTORY_SIZE = 100
 
 
 class SimConnection(IConnection):
-    def _rocket_handshake(self):
-        assert self.rocket.stdout.read(3) == b"SYN"
-        # Uncomment for FW debuggers, for a chance to attach debugger
-        # input("Recieved rocket SYN; press enter to respond with ACK and continue\n")
-        self.rocket.stdin.write(b"ACK")
-        self.rocket.stdin.flush()
-
     def __init__(self, firmwareDir, executableName):
         self.executablePath = os.path.join(firmwareDir, executableName)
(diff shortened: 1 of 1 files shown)


SimConnection.py
```diff
@@ -19,6 +19,13 @@ class SimPacketId(Enum):
 
 
 class SimConnection(IConnection):
+    def _rocket_handshake(self):
+        assert self.rocket.stdout.read(3) == b"SYN"
```
SilentSys, line 23: Just for consistency's sake, let's use `self.stdout` instead of `self.rocket.stdout`. Unless there's a reason for doing it this way that I'm not aware of.

SimConnection.py
```diff
@@ -19,6 +19,13 @@ class SimPacketId(Enum):
 
 
 class SimConnection(IConnection):
+    def _rocket_handshake(self):
```
SilentSys, line 22: Can we please move this somewhere below the constructor.

SimConnection.py
jesseli2002: Not actually sure what StreamLogger does so I opted to not touch it. Only reason I'd note is to keep self._rocket_handshake() as close to Popen as possible, which would put it in front of the setting up of self.stdout

SilentSys: It has a circular buffer of the read history. It can be very helpful for debugging.

SilentSys: Especially now that you moved the handshake call below the creation of the StreamLogger, we should use it lol

PR #16, opened 2020-08-02
Fix MapData name overlap

ADSteele916:
Just a small naming change to address a TODO in MapData.py. Changed file name to map_data.py, class name to MapData, and refactored references to each. This way we can avoid appending the redundant "Class" to the class's name and be a bit more in line with PEP 8 conventions.


ADSteele916:
No worries. I just spotted it and figured it would take 5 minutes to fix, so I figured I might as well.

PR #17, opened 2020-08-03
SIM fixes

SilentSys:
- Made the GS not crash when an unimplemented packet is received
- Cleaned up StreamLogger (ugh generators)

Thanks Jesse

### DIFF
--- a/RadioController.py
+++ b/RadioController.py
@@ -39,6 +39,7 @@ for i in SubpacketIDs.get_list_of_sensor_IDs():
 
 
 class RadioController:
+
     def __init__(self, bigEndianInts, bigEndianFloats):
         """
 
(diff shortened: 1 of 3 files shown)


StreamLogger.py
```diff
@@ -1,93 +1,71 @@
...
+    def __init__(self, bufstream, size) -> None:
         """
-
-        :param stream:
-        :type stream:
-        :param size:
+        :param bufstream: Bufferred stream like a subprocess stdout, that supports read() and peek().
+        :type bufstream:
+        :param size: Size of in-memory circular log buffer
         :type size:
```
ADSteele916, line 16: Types for parameters?

RadioController.py
```diff
@@ -9,7 +9,7 @@
 # TODO Review these based on spec
 # Map subpacket id (int) to length (int) in bytes. Only includes types with CONSTANT lengths
 PACKET_ID_TO_CONST_LENGTH: Dict[int, int] = {
-    SubpacketEnum.STATUS_PING.value: 4,
+    SubpacketEnum.STATUS_PING.value: 10,
     SubpacketEnum.EVENT.value: 2,
     SubpacketEnum.GPS.value: 25,
     SubpacketEnum.ACKNOWLEDGEMENT.value: 0000,  # TODO ack length?
```
DSep, line 15: These will be corrected in mine as well! Not correct at the moment.

PR #20, opened 2020-08-10
Implementing subpacket parsers

DSep:


RocketData.py
```diff
@@ -39,28 +39,27 @@
 #
```
DSep, line 39: Need to cleanup this files comments and todos, since all packets will have timestamps

RadioController.py
```diff
@@ -1,43 +1,47 @@
...
+#     SubpacketEnum.STATUS_PING.value: 5,
+#     SubpacketEnum.MESSAGE.value: 6,
+#     SubpacketEnum.EVENT.value: 5,
+#     SubpacketEnum.CONFIG.value: 5,
+#     SubpacketEnum.GPS.value: 5,
+#     SubpacketEnum.ACKNOWLEDGEMENT.value: ,???
+#     SubpacketEnum.BULK_SENSOR.value: 5,
+# }
+# for i in SubpacketIDs.get_list_of_sensor_IDs():
+#     PACKET_ID_TO_HEADER_SIZE[i] = 1
```
DSep, line 41: Let me know what you think, @jesseli2002 especially, regarding specifying header size somewhere or not

DebugConnection.py
```diff
@@ -24,13 +26,15 @@ def _run(self) -> None:
...
             with self.lock:
                 if not self.callback:
                     continue
 
-                bulk_sensor_arr: bytearray = self.bulk_sensor_mock_random()
-                self.callback(bulk_sensor_arr)
+                # bulk_sensor_arr: bytearray = self.bulk_sensor_mock_random()
+                # statusPingArr: bytearray = self.statusPingMockSetValues()
+                messageArr: bytearray = self.messageValues()
+                self.callback(messageArr)
```
ADSteele916, line 37: DebugConnection doesn't seem to work anymore. All of the sensor values just become None (or stay 0) after ~5 seconds.

ReadThread.py
```diff
@@ -48,19 +48,21 @@ def run(self):
 
             data = self.dataQueue.get(block=True, timeout=None)  # Block until something new
             self.dataQueue.task_done()
-
-            byteList = [[x] for x in data]  # sad that we still need this
+            print(data)
```
ADSteele916, line 51: Is this necessary? I feel like the console gets more than enough usage already.

RadioController.py
```diff
@@ -101,157 +109,302 @@ def parse_data(self, type_id, byte_list, length):
...
+
+    # returns the current value then increments
+    def currAndInc(self, interval=9):
+        num = self.num
+        if interval == 0:
+            self.num += self.interval
+        else:
+            self.num += interval
+
+        return num
```
ADSteele916, line 403: Definitely belongs in detail.py or something similar, IMO.

RadioController.py
```diff
@@ -101,157 +109,302 @@ def parse_data(self, type_id, byte_list, length):
...
-    def gps(self, byte_list, length):  # TODO
+    def gps(self, byte_list, **kwargs):
         """
 
         :param byte_list:
         :type byte_list:
-        :param length:
-        :type length:
+        :param **kwargs:
+        :type **kwargs:
```
ADSteele916, line 272: Not entirely sure if this is how we're supposed to do documentation for *args and **kwargs. Might be worth double-checking. Same goes for other instances of this.

DebugConnection.py
SilentSys: Good catch! Manual CI lol

ReadThread.py
DSep: Nope. To be removed.

RadioController.py
```diff
@@ -101,157 +109,302 @@ def parse_data(self, type_id, byte_list, length):
...
+        data[SubpacketEnum.ACCELERATION_X.value] = self.fourtofloat(byte_list[curr_byte.curr():curr_byte.next(4)])
+        data[SubpacketEnum.ACCELERATION_Y.value] = self.fourtofloat(byte_list[curr_byte.curr():curr_byte.next(4)])
+        data[SubpacketEnum.ACCELERATION_Z.value] = self.fourtofloat(byte_list[curr_byte.curr():curr_byte.next(4)])
+        data[SubpacketEnum.ORIENTATION_1.value] = self.fourtofloat(byte_list[curr_byte.curr():curr_byte.next(4)])
+        data[SubpacketEnum.ORIENTATION_2.value] = self.fourtofloat(byte_list[curr_byte.curr():curr_byte.next(4)])
+        data[SubpacketEnum.ORIENTATION_3.value] = self.fourtofloat(byte_list[curr_byte.curr():curr_byte.next(4)])
+        data[SubpacketEnum.LATITUDE.value] = self.fourtofloat(byte_list[curr_byte.curr():curr_byte.next(4)])  # TODO Should lat+lon be doubles??
+        data[SubpacketEnum.LONGITUDE.value] = self.fourtofloat(byte_list[curr_byte.curr():curr_byte.next(4)])
+        data[SubpacketEnum.STATE.value] = byte_list[curr_byte.currAndInc(1)]
```
SilentSys, line 313: Starting to think that we should be using [io.BytesIO](https://docs.python.org/3/library/io.html#io.BytesIO). Note that it inherits [io.BufferedIOBase](https://docs.python.org/3/library/io.html#io.BufferedIOBase), so it has all the cool stream methods like `read()`, `write()`, `peak()` etc.

This would simplify to:
```suggestion
        data[SubpacketEnum.CALCULATED_ALTITUDE.value] = self.fourtofloat(byte_list.read(4))  # TODO Double check it is calculated barometer altitude with firmware
        data[SubpacketEnum.ACCELERATION_X.value] = self.fourtofloat(byte_list.read(4))
        data[SubpacketEnum.ACCELERATION_Y.value] = self.fourtofloat(byte_list.read(4))
        data[SubpacketEnum.ACCELERATION_Z.value] = self.fourtofloat(byte_list.read(4))
        data[SubpacketEnum.ORIENTATION_1.value] = self.fourtofloat(byte_list.read(4))
        data[SubpacketEnum.ORIENTATION_2.value] = self.fourtofloat(byte_list.read(4))
        data[SubpacketEnum.ORIENTATION_3.value] = self.fourtofloat(byte_list.read(4))
        data[SubpacketEnum.LATITUDE.value] = self.fourtofloat(byte_list.read(4))  # TODO Should lat+lon be doubles??
        data[SubpacketEnum.LONGITUDE.value] = self.fourtofloat(byte_list.read(4))
        data[SubpacketEnum.STATE.value] = byte_list[curr_byte.read(1)]
```

And we wouldn't need the `Count` helper class either. Also wouldn't need to pass around the header length.

What do you think about it?

RadioController.py
```diff
@@ -101,157 +109,302 @@ def parse_data(self, type_id, byte_list, length):
...
+        BAROMETER = 'BAROMETER'
+        GPS = 'GPS'
+        ACCELEROMETER = 'ACCELEROMETER'
+        TEMPERATURE = 'TEMPERATURE'
+        IMU = 'IMU'
+        SENSOR_TYPES = [OVERALL_STATUS, BAROMETER, GPS, ACCELEROMETER, IMU, TEMPERATURE]
+        DROGUE_IGNITER_CONTINUITY = 'DROGUE_IGNITER_CONTINUITY'
+        MAIN_IGNITER_CONTINUITY = 'MAIN_IGNITER_CONTINUITY'
+        FILE_OPEN_SUCCESS = 'FILE_OPEN_SUCCESS'
+        OTHER_STATUS_TYPES = [DROGUE_IGNITER_CONTINUITY, MAIN_IGNITER_CONTINUITY, FILE_OPEN_SUCCESS]
```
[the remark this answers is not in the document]
SilentSys: I think we spoke about this refactoring before. If you feel like its beyond the scope of this ticket, we can make a new ticket for it and do it whenever.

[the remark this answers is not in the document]
DSep: I think it would be good to separate, just to have a separate PR and discussion and to let the deflattening branch have a chance to get merged first.

RadioController.py
DSep: Moved. Less obvious is bitFromByte. I'm leaving it in for now

DebugConnection.py
DSep: I've been using it for testing and I'm not sure how's best to get all the info.
Let's discuss this.

RadioController.py
DSep: Tried something out after searching, from https://stackoverflow.com/questions/47617701/python-how-to-docstring-kwargs-and-their-expected-types. Works nice in docstring format I think, and intellisense. Lmk if you prefer another way.

ADSteele916: No, that's fine by me. Just wanted to make sure we're not making up our own standard.

ADSteele916 (approved):
LGTM

SilentSys (approved):
LGTM!

PR #21, opened 2020-08-11
Add rocket profiles

ADSteele916:
Added rocket profile system so the ground station works with multiple rockets.


rocket_profile.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class RocketProfile:
+    def __init__(self, buttons: Dict[str, str], labels: List[Label]):
+        self.buttons = buttons
+        self.labels = labels
+
+    def update_labels(self, main_window) -> None:
+        for label in self.labels:
+            exec(f"main_window.{label.name}Label.setText(label.update(main_window))")
```
SilentSys, line 13:
```suggestion
            getattr(main_window, label.name + "Label").setText(label.update(main_window))
```

Would something like this work? This might help get rid of the exec, make it more debuggable and understandable by the IDE?

main.py
```diff
@@ -93,6 +96,65 @@ def __init__(self, connection) -> None:
...
+            exec(f"self.{button}Button.setText(QtCore.QCoreApplication.translate('MainWindow', '{button}'))")
+
+        for button, command in self.rocket.buttons.items():
+            exec(f"self.{button}Button.clicked.connect(lambda _: self.sendCommand('{command}'))", {"self": self})
+
+    def setup_labels(self):
+        row = 0
+        for label in self.rocket.labels:
+            name = label.name
+            exec(f"self.{name}Text = QtWidgets.QLabel(self.centralwidget)")
```
SilentSys, line 140:
```suggestion
            setattr(self, name + "Text", QtWidgets.QLabel(self.centralwidget))
```
I think we can use `setattr` here.

main.py
```diff
@@ -93,6 +96,65 @@ def __init__(self, connection) -> None:
...
+            sizePolicy.setHorizontalStretch(0)
+            sizePolicy.setVerticalStretch(0)
+            exec(f"sizePolicy.setHeightForWidth(self.{name}Text.sizePolicy().hasHeightForWidth())")
+            exec(f"self.{name}Text.setSizePolicy(sizePolicy)")
+            exec(f"self.{name}Text.setObjectName('{name}Text')")
+            exec(f"self.gridLayout_6.addWidget(self.{name}Text, {row}, 0, 1, 1)")
+
+            exec(f"self.{name}Label = QtWidgets.QLabel(self.centralwidget)")
+            exec(f"self.{name}Label.setObjectName('{name}Label')")
+            exec(f"self.gridLayout_6.addWidget(self.{name}Label, {row}, 1, 1, 1)")
```
SilentSys, line 151: I think you can use `setattr` or `getattr` for these too. Possibly everywhere else too actually.

rocket_profile.py
ADSteele916: I'll try that out. Thanks!

main.py
```diff
@@ -108,18 +108,19 @@ def setup_buttons(self):
...
+            qt_button.setSizePolicy(sizePolicy)
             font = QtGui.QFont()
             font.setPointSize(35)
             font.setKerning(True)
-            exec(f"self.{button}Button.setFont(font)")
-            exec(f"self.{button}Button.setObjectName('{button}Button')")
-            exec(f"self.gridLayout_5.addWidget(self.{button}Button, {row}, {col}, 1, 1)")
+            qt_button.setFont(font)
+            qt_button.setObjectName(f"{button}Button")
+            self.gridLayout_5.addWidget(qt_button, row, col, 1, 1)
```
SilentSys: Wow looks great! No execs and its all actual code 🚀 

I'm assuming it all works just fine?

main.py
```diff
@@ -93,6 +96,74 @@ def __init__(self, connection) -> None:
...
+
+        grid_width = math.ceil(math.sqrt(len(self.rocket.buttons)))
+
+        # Trying to replicate PyQt's generated code from a .ui file as closely as possible. Hence the execs and custom
+        # button object names.
+        row = 0
+        col = 0
+        for button in self.rocket.buttons.keys():
+            setattr(self, button + "Button", QtWidgets.QPushButton(self.centralwidget))
+            qt_button = getattr(self, button + "Button")
```
SilentSys, line 112:
```suggestion
            qt_button  = QtWidgets.QPushButton(self.centralwidget)
            setattr(self, button + "Button", qt_button)
```

Does this work?

main.py
ADSteele916: Everything works as before.

main.py
ADSteele916: Yes. Good catch.

rocket_profile.py
```diff
@@ -0,0 +1,23 @@
...
+    def update_labels(self, main_window: "MainApp") -> None:
+        for label in self.labels:
+            getattr(main_window, label.name + "Label").setText(label.update(main_window))
+
+
+tantalus = RocketProfile({"Arm": "arm", "Status": "status"}, default_labels)
+co_pilot = RocketProfile(
+    {"Arm": "arm", "Halo": "halo", "Data": "data", "Status": "status"},
+    default_labels + [Label("Test", lambda main_window: "Hey")],
+)
```
ADSteele916, line 23: These are just dummy profiles for the moment. I'd personally like to change these to whatever buttons and labels the rockets in question would actually use (so we don't have a label that just says "Hey" in master), but if we don't know what the actual rocket profiles will look like yet. this is fine for now.

SilentSys: You could add a "test separation" button to tantalus and "tank pressure", "tank temp", "chamber pressure", "chamber temp" to co-pilot.

ADSteele916: Sounds promising. And what would their update methods look like?

ADSteele916: Sounds promising. And what would their update methods look like?

ADSteele916: Sounds promising. And what would their update methods look like?

SilentSys: Dont really have that functionality yet. You could add a no-op that makes it clear it is a no-op/unimplemented.

rocket_profile.py
```diff
@@ -0,0 +1,24 @@
...
+        for label in self.labels:
+            getattr(main_window, label.name + "Label").setText(
+                label.update(main_window)
+            )
+
+
+tantalus = RocketProfile({"Arm": "arm", "Status": "status"}, tantalus_labels)
+co_pilot = RocketProfile(
+    {"Arm": "arm", "Halo": "halo", "Data": "data", "Status": "status"}, cp_labels
+)
```
SilentSys, line 24: Could we please move each of these profiles into their own files? In the future, rocket specific behavior might get a lot more complicated, we might have different connections for different rockets, different protocols, maybe we'll have physics models, etc. By having them in separate files, that would allow us to have one directory for each rocket containing all the rocket-specific files.

label.py
```diff
@@ -0,0 +1,93 @@
...
+
+tantalus_labels = default_labels + [
+    Label("TestSeparation", update_test_separation, "Test Separation")
+]
+
+cp_labels = default_labels + [
+    Label("TankPressure", update_tank_pressure, "Tank Pressure"),
+    Label("ChamberPressure", update_chamber_pressure, "Chamber Pressure"),
+    Label("ChamberTemp", update_chamber_temp, "Chamber Temperature"),
+]
```
SilentSys, line 93: All the functions are nice to have here, but can we please move the lists into the rocket-specific files mentioned in the other comment? Having one file that describes everything for each rocket is nice because when someone eventually needs to change something they don't have to go digging through multiple files to find where it came from. 

Also, we can get rid of the `default_labels` list as it might not be very default for long and it will help with keeping everything for each rocket together.

main.py
```diff
@@ -93,6 +96,74 @@ def __init__(self, connection) -> None:
...
 
+    def setup_buttons(self):
+        """
+
+        """
+
+        grid_width = math.ceil(math.sqrt(len(self.rocket.buttons)))
+
+        # Trying to replicate PyQt's generated code from a .ui file as closely as possible. Hence the execs and custom
+        # button object names.
```
SilentSys, line 107: Nice job with the reverse-engineering and replication of PyQt btw!

label.py
ADSteele916: Done. The root directory of the repository is getting a bit hectic though. I'll likely submit another pull request that puts all of my new files into a "profiles" folder or something and do the same for some other files once this branch (and possibly Sepand's) gets pulled.

SilentSys: That would be awesome if you do that. I see you already added yourself to https://trello.com/c/VY5osi76/230-reorganize-ground-station-directories , so I guess this fits within its scope nicely.

PR #24, opened 2020-08-23
Fix MapBox TileGrid cache naming

ADSteele916:
- Previously, TIleGrids would use their defining coordinates to generate a file name. However, since the coordinates are floats, this means that unless the latitude and longitude haven't updated yet, a cached version will not be used, even if it is identical to the one wanted. I changed the naming scheme so that it now relies on the min and max tile x and tile y values (and the scale). This should allow duplicate maps to be reused.
- Updated test_MapBox to reflect the new changes.


ADSteele916:
> LGTM! This should reduce the number of files saved to disk too right?

Yes, because now cached TileGrid images will actually get used even if the coordinates change slightly.

PR #26, opened 2020-09-06
Quick fix for com_window.main

ADSteele916:
- Removed the if __name__ == "__main__" statement in com_window.main, as it does not work and is no longer necessary.
- Minor fix for rocket_data.

### DIFF
--- a/com_window/main.py
+++ b/com_window/main.py
@@ -1,5 +1,4 @@
 import os
-import sys
 
 import PyQt5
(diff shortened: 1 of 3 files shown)


SilentSys:
Is the com_window directory also supposed to have an init file? Its the only dir that doesnt have it I think.

SilentSys (approved):
Wow that was fast. LGTM!

PR #27, opened 2020-09-06
HW simulating

jesseli2002:


SilentSys (changes requested):

profiles/rocket_profile.py
```diff
@@ -1,15 +1,21 @@
...
 class RocketProfile:
-    def __init__(self, buttons: Dict[str, str], labels: List[Label]):
+    def __init__(
+        self, buttons: Dict[str, str], labels: List[Label], hw_sim: HWSim = None
+    ):
         self.buttons = buttons
         self.labels = labels
+        if not hw_sim:
+            hw_sim = HWSim()
+        self.hw_sim = hw_sim
```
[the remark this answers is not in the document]
jesseli2002: Moved constructor of HWSim into SimConnectionFactory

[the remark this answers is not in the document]
SilentSys: Nice solution. I dig it.

connections/sim/hw_sim.py
```diff
@@ -0,0 +1,89 @@
...
+        for (test, read, fire), broke in zip(ignitors, broken):
+            ign = IgnitorSim(broke)
+            self.ignitors.append(ign)
+            self.ignitor_tests[test] = ign
+            self.ignitor_reads[read] = ign
+            self.ignitor_fires[fire] = ign
+
+    def digital_write(self, pin, val):
+        """
+        :param pin: Should be a test pin.
```
SilentSys, line 74: Should we maybe assert, throw, or warn if its not in any of the lists? Always good to fail noisily, might save us some day.

connections/sim/sim_connection.py
```diff
@@ -107,20 +111,29 @@ def _handleBuzzer(self):
...
         length = self._getLength()
         data = self.stdout.read(length)
         self._xbee.recieved_from_rocket(data)
 
+    def _handleAnalogRead(self):
+        length = self._getLength()
+        assert length == 1
+        pin = self.stdout.read(length)[0]
+        result = self._hw_sim.analog_read(pin).to_bytes(2, "big")
+        self._send_sim_packet(b"a", result)
```
SilentSys, line 129: Can we please make the `b"a"` a constant so that people don't get confused about the magic value?

connections/sim/sim_connection.py
```diff
@@ -43,26 +46,27 @@ def __init__(self, firmwareDir, executableName):
...
-        packet += len(data).to_bytes(length=2, byteorder="big")
-        packet += data
-
+    def _send_sim_packet(self, id_, data):
+        packet = id_ + len(data).to_bytes(length=2, byteorder="big") + data
         for b in packet:  # Work around for windows turning LF to CRLF
             self.rocket.stdin.write(bytes([b]))
         self.rocket.stdin.flush()
 
+    _send_radio_sim = partialmethod(_send_sim_packet, b"R")
```
SilentSys, line 68: For consistency, clarity, and simplicity, can we please write this out as a regular function?

```suggestion
    def _send_radio_sim(self, data):
        _send_sim_packet(b"R", data)
```

SilentSys, line 68: Also lets make the `b"R"` a constant as well.

connections/sim/sim_connection.py
```diff
@@ -43,26 +46,27 @@ def __init__(self, firmwareDir, executableName):
...
     def send(self, data):
         self._xbee.send_to_rocket(data)
 
-    def _send_radio_sim(self, data):
-        packet = b"R"
-        packet += len(data).to_bytes(length=2, byteorder="big")
-        packet += data
-
+    def _send_sim_packet(self, id_, data):
+        packet = id_ + len(data).to_bytes(length=2, byteorder="big") + data
```
SilentSys, line 63: Does the underscore in `id_` mean anything? Also if we cast the id to bytes in this function, we don't have to do `b"x"` everywhere.

profiles/rockets/tantalus.py
```diff
@@ -13,4 +21,21 @@
...
-tantalus = RocketProfile({"Arm": "arm", "Status": "status"}, tantalus_labels)
+_MAIN_IGN_TEST = 4
+_MAIN_IGN_READ = 14
+_MAIN_IGN_FIRE = 16
+_DROGUE_IGN_TEST = 17
+_DROGUE_IGN_READ = 34
+_DROGUE_IGN_FIRE = 35
+
+
+_hw_sim_dat = [
```
SilentSys, line 32:
```suggestion
_hw_sim_ignitors = [
```

Seems less vague.

SilentSys (changes requested):
Just some small things. Overall, love the precedent this sets for hardware simulation and testing.

connections/sim/sim_connection.py
jesseli2002: id is a built-in function, and Python style guide suggests appending an underscore to avoid naming conflicts with builtins.

connections/sim/hw_sim.py
jesseli2002: Note that not all output pins are being simulated (yet) - e.g. some are connected to status LEDs, one is connected to a buzzer and is toggled at different frequencies to make different tones. IIRC I connected SIM directly to HW_Sim, so you'd have to define all of the valid output pins (as opposed to just the ones we're simulating).

connections/sim/sim_connection.py
jesseli2002: but functional programming.........
at least check that the usage doesn't go wonky (`partialmethod` I know just makes things work), and if you're doing this then the functools import can be removed too

ADSteele916 (approved):
LGTM once Andy's changes are in.

SilentSys (approved):

PR #29, opened 2020-10-05
Changed imresize to pillow.resize

akoen:
image

This should be the correct fix, but both master and my branch fail the above test. Since I don't have a passing implementation to compare to, I'd appreciate it if someone could run my code on their end.

### DIFF
--- a/main_window/mapping/mapping_thread.py
+++ b/main_window/mapping/mapping_thread.py
@@ -5,7 +5,8 @@ from multiprocessing import Process, Queue
 
 from PyQt5 import QtCore
 from PyQt5.QtCore import pyqtSignal
-from scipy.misc import imresize
+import numpy as np
+from PIL import Image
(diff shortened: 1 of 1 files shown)


SilentSys:
@ADSteele916 @ADungate Why do I vaguely remember us already making these changes? I'm spooked now at how they didnt make it in.

SilentSys:
Ah, it was #22, it was done for the marker and the map was missed.

SilentSys (approved):
The tests pass on my machine. Weird that it doesn't pass on yours. Did you set up your mapbox api key?

Thanks for fixing this 🚀

akoen:
Yes, I believe so. If I can see the map, does that mean that my API key is working?

Will test again in the coming days.

SilentSys:
Yeah, that means your key is working.

PR #30, opened 2020-10-09
Fixed windows CR LF filter

SilentSys:


connections/sim/stream_logger.py
```diff
@@ -42,7 +42,7 @@ def _read_and_log(self, stream_gen):
     def _crcrlfFilter(self, stream):
         while True:
             c = stream.read(1)
-            if c == CR and stream.peek(1)[0] == LF:
+            if c[0] == CR and stream.peek(1)[0] == LF:
```
SilentSys, line 45: This is actually the entire fix. The rest is just misc stuff.

jesseli2002: Fix aside, I'm curious how this ever worked, if at all

jesseli2002 (approved):

connections/sim/stream_logger.py
SilentSys: Not sure, but it will never happen again (hopefully (see new commit))

DSep (approved):
Sweet; looking forward to trying this today on our call and reporting what the state of SIM is

PR #32, opened 2020-10-17
Use global logger with log levels and saving to file

SilentSys:
Swapped out print statements with Python's built-in logging module.

This allows us to:

- Easily log and prints at the same time
- Easily print exception info from an except block
- Include useful information in each print (time, file, function, etc.)
- Assign log-levels and filter console based on them (maybe release builds wont include messy debug logs?)
- Create custom handlers -  maybe we could make the UI console be a logging handler and it can print high level logs? Not sure about qt thread safety with this though

https://trello.com/c/7apMC3Ym/


ADSteele916 (approved):
LGTM

DSep (approved):
Looks great! So nice to have proper logging 😄

PR #33, opened 2020-10-18
Make Sim search multiple dirs for build file

alxshev:
Code now checks multiple folders for build file (UBCRocketStation/FW/rocket, and FLARE/avionics/build/<<something which contains either copilot or tantalus>>).


ADSteele916 (dismissed):
LGTM

main_window/mapping/mapping_thread.py
```diff
@@ -5,8 +5,7 @@
 
 from PyQt5 import QtCore
 from PyQt5.QtCore import pyqtSignal
-import numpy as np
-from PIL import Image
+# from scipy.misc import imresize
```
DSep, line 8: Can remove the artifact I think

DSep (approved):
Looks good!

connections/sim/sim_connection_factory.py
```diff
@@ -20,18 +19,41 @@ def requiresComPort(self) -> bool:
     def requiresBaudRate(self) -> bool:
         return False
 
+    def _getParentDir(self, dir):
+        dirs = dir.split("/")
+        return "/".join(dirs[0:len(dirs) - 1])
```
SilentSys, line 24: This will likely not work on Windows due to it using backslashes instead.

Maybe we can use this Python built-in solution: https://stackoverflow.com/questions/2860153/how-do-i-get-the-parent-directory-in-python

connections/sim/sim_connection_factory.py
```diff
@@ -20,18 +19,41 @@ def requiresComPort(self) -> bool:
...
+            comPort: Union[int, str, None] = None,
+            baudRate: Union[int, None] = None,
+            rocket=None,
     ):
         assert rocket is not None
-        executableName = EXECUTABLE_NAME
+
+        # Check FW (child) dir and FLARE (neighbour) dir for rocket build files
+        # If multiple build files found throw exception
+        flarePath = self._getParentDir(LOCAL) + "/FLARE/avionics/build/"
```
SilentSys, line 36: Take a look at `os.path.join()`

SilentSys (changes requested):
Looks good to me! Just some suggestions about paths for Windows.

connections/sim/sim_connection_factory.py
jesseli2002: It may just work, actually - Windows will let you use forward slashes and resolves them to backslashes.
That being said, the most Pythonic way of going about this I think would be through the high level `Pathlib` library which provides object oriented filesystem manipulation. (I've never gotten the hang of it though and usually still use lower level `os.path.join` etc., directly manipulating strings)

connections/sim/sim_connection_factory.py
```diff
@@ -20,18 +19,41 @@ def requiresComPort(self) -> bool:
...
+        neighbourBuildFiles = [s for s in os.listdir(flarePath) if rocket.rocket_name.lower() in s.lower()]
+
+        childBuildFileExists = "program" in os.listdir('./FW')
+        neighbourBuildFileExists = bool(neighbourBuildFiles)
+
+        if childBuildFileExists and neighbourBuildFileExists:
+            raise Exception(
```
jesseli2002, line 43: If you wanted to be fancy you could create a new Exception class for firmware errors or something.

connections/sim/sim_connection_factory.py
SilentSys: The join will work. The split won't.

main_window/mapping/mapping_thread.py
alxshev: ok

connections/sim/sim_connection_factory.py
alxshev: updated it

connections/sim/sim_connection_factory.py
alxshev: Yeah I made it use os.path.join now

connections/sim/sim_connection_factory.py
alxshev: Yeah it uses pathlib now

main_window/mapping/mapping_thread.py
SilentSys, line 8:
```
[2020-10-29 10:27:45,974] (ERROR) mapping_thread.py:processMap: Exception in processMap process
Traceback (most recent call last):
  File "B:\Documents\CODE\UBCRocketGroundStation\main_window\mapping\mapping_thread.py", line 220, in processMap
    resizedMapImage = imresize(largeMapImage, (
NameError: name 'imresize' is not defined
```

Will need to uncomment that import.

connections/sim/sim_connection_factory.py
```diff
@@ -21,17 +22,37 @@ def requiresBaudRate(self) -> bool:
...
+            baudRate: Union[int, None] = None,
+            rocket=None,
     ):
         assert rocket is not None
-        executableName = EXECUTABLE_NAME
+
+        # Check FW (child) dir and FLARE (neighbour) dir for rocket build files
+        # If multiple build files found throw exception
+        flarePath = os.path.join(Path(LOCAL).parent, 'FLARE', 'avionics', 'build')
+        localPath = os.path.join(os.curdir, 'FW')
```
SilentSys, line 35:
```
Traceback (most recent call last):
  File "B:\Documents\CODE\UBCRocketGroundStation\com_window\main.py", line 68, in doneButtonPressed
    rocket=rocket,
  File "B:\Documents\CODE\UBCRocketGroundStation\connections\sim\sim_connection_factory.py", line 40, in construct
    childBuildFileExists = "program" in os.listdir(localPath)
FileNotFoundError: [WinError 3] The system cannot find the path specified: '.\\FW'
```
Can we please use `LOCAL` for this one too. `os.curdir` is just a relative path and wont give the correct result if the program is started from a different directory. Might also have issues with PyInstaller.

SilentSys (changes requested):

main_window/mapping/mapping_thread.py
alxshev:
```suggestion
from scipy.misc import imresize
```

connections/sim/sim_connection_factory.py
alxshev:
```suggestion
        localPath = os.path.join(LOCAL, 'FW')
```

alxshev (commented):
revised path manipulations

SilentSys (commented):
I undid the changes in mappying_thread.py as I just realized that was some sort of collision/conflict with #29

SilentSys (approved):
LGTM

ADSteele916 (approved):
LGTM

PR #35, opened 2020-10-19
Event stats, radio packet integration tests with debug connection

SilentSys:
- Implemented event_stats module and used it in all tests that depended on sleeps
     - event_stats will likely be very useful in every integration test we write as it allows us to wait until threads process data before asserting on their results
- Radio packet integration tests with debug connection


DSep (approved):
Looks good, but to properly understand I think I'll need a run down of the new debug system next meeting.
Really nice to have those debug packet creators extracted like that. I hope it'll be okay (design-wise) to use them in SIM tests potentially too?

And if I'm not mistaken the events appear to be global objects following a different design pattern than usual, but are nicely done and very convenient. I'll have to hear the design process and reasoning next meeting.

SilentSys:
@DSep Sounds good, look forward to chatting about it next meeting.

DSep (approved):
LGTM! Thanks for explaining the pieces during the call. It's pretty cool stuff! 
And nice functional moves Alex 😄

ADSteele916 (approved):
LGTM

PR #36, opened 2020-10-29
Fixed pyinstaller builds, replaced mk.bat with build.py

SilentSys:
- Fixed PyInstaller builds
     - Added CI "self-test" step to ensure it never breaks again
- Replaced mk.bat with OS-agnostic build.py
     - Sets up venv, builds executable, runs self-test
     - Documented in http://confluence.ubcrocket.com/display/AV/build.py+Usage (GS setup guide will be updated after merge)


util/self_test.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def __init__(self):
+        self.thread = threading.Thread(target=self._run_self_test, daemon=True, name="SelfTestThread")
+
+    def start(self):
+        self.thread.start()
+
+    def _run_self_test(self):
+        try:
+            LOGGER.warning("SELF TEST STARTED")
```
jesseli2002, line 24: I would've thought that warnings would be reserved for things that are probably bad...

SilentSys: It's bad if you see this log when you're not expecting it. You're right, I know what you're saying, but if we ever filter our logs/console by log level, we might want this to still shows up.

I'd be happy to demote it to an info log if you think that's more appropriate.

start.py
```diff
@@ -1,10 +1,46 @@
 import sys
-
+import multiprocessing
+import argparse
 from com_window.main import comWindow
 from PyQt5 import QtWidgets
```
DSep, line 5:
```suggestion
from PyQt5 import QtWidgets
if getattr(sys, 'frozen', False) and sys.platform == 'darwin':
    matplotlib.use('Qt5Agg')
```

Can also be the TkAgg backend

main_window/mapping/mapbox_utils.py
```diff
@@ -9,11 +9,11 @@
 import numpy as np
 from matplotlib import pyplot as plt
```
DSep, line 10:
```suggestion
if getattr(sys, 'frozen', False) and platform == 'darwin':
    matplotlib.use('Qt5Agg')
from matplotlib import pyplot as plt
```
Can also be the TkAgg backend

DSep (changes requested):
Nice!
As discussed, on macOS it needs the user's python build to be a framework build. Not really avoidable, and the user will need to rebuild with the right flags if necessary.

Also as discussed, there's also an issue with the matplotlib backend on macOS. It would be convenient to have the matplotlib backend configured to something that works outside of pycharm, when a user is running build.py. (Check out the suggestions of 2 possible places where the override for macOS can work)

DSep (dismissed):
LGTM!

SilentSys:
[start.py argparse com_window skip](https://trello.com/c/BuWXoSs4/256-startpy-argparse-comwindow-skip)

SilentSys:
[Fix PyInstaller'd GS not working and add to CI](https://trello.com/c/saVRey8g/184-fix-pyinstallerd-gs-not-working-and-add-to-ci)

SilentSys:
[Create mk.py](https://trello.com/c/DxH1580h/185-create-mkpy)

akoen:
Running `python ./build.py --skip test` on Linux in a fresh directory gives me the following error:

image

To fix the issue, the command `env PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install 3.7.0 ` needs to be run beforehand. This is documented here https://github.com/pyenv/pyenv/wiki#how-to-build-cpython-with---enable-shared.

I have updated the README.

SilentSys:
Sweet, thanks @akoen !

DSep:
Quoting: https://github.com/UBC-Rocket/UBCRocketGroundStation/pull/36#pullrequestreview-522096368
> As discussed, on macOS it needs the user's python build to be a framework build. Not really avoidable, and the user will need to rebuild with the right flags if necessary.

@SilentSys @akoen That is nearly identical to what was needed needed with macOS + pyenv btw, in case we want it as a unix-based instruction instead, with:
- macOS: env PYTHON_CONFIGURE_OPTS="--enable-framework" pyenv install 3.7.3
- Linux: env PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install 3.7.3

PR #38, opened 2020-11-06
Profile refactoring, HW Sim refactoring, SIM CI Test

SilentSys:
The scope of the refactoring got a little bigger than I planned. Key notes:

- Used abstract class for rocket profile
     - enforces read-only properties
     - factory pattern for things that should only be constructed if needed (HW Sim)
- Did a bunch of refactoring in HW Sim
     - Together with the rocket profile refactoring, this should let us just list our all the hardware in the profile in a intuitive way
- Created first SIM Integration test
     - Enabled in CI
     - There is currently a bug where we're getting occasional bad sensor values over SIM, so currently not asserting on value, just that we got something
     - This should lay out the foundation/framework for all future SIM tests


SilentSys:
[SIM CI with both repos](https://trello.com/c/PwKlNUWq/255-sim-ci-with-both-repos)

DSep (approved):
So many things! LGTM 👍

SilentSys:
[Kill all threads and processes at the end of each integration test](https://trello.com/c/TvuyeVb4/261-kill-all-threads-and-processes-at-the-end-of-each-integration-test)

PR #45, opened 2020-11-20
Print info level logs and above to UI

SilentSys:
image

Looks bad if you don't resize the window though. we should probably address our layout soon.


SilentSys:
[Implement printing to UI from RocketData for messages and statuses](https://trello.com/c/IOHIEH8U/224-implement-printing-to-ui-from-rocketdata-for-messages-and-statuses)

DSep (approved):
🙏

PR #46, opened 2020-11-25
Gs clock

akoen:
image

Let me me know if you'd like me to remove the logging: it causes a lot of spam but at least looks pretty.


akoen:
Also it still needs tests (which I'm happy to write if this approach looks good).

SilentSys (changes requested):
Looks great!

Yeah, lets just add some tests. I'm thinking some unit-tests to test the clock logic in `test_hw_sim.py`. And then maybe a `test_sim.py` test that increases the clock time by some very large amount and then asserts that `main_app.rocket_data.lastvalue(SubpacketEnum.TIME.value)` changed by at least that amount (compared to an initial value).

Can't think of any other good integration tests for this off the top of my head but if you can, please add them as well.

SilentSys:
Just as mentioned in https://github.com/UBC-Rocket/FLARE/pull/132, you'll need to update `required_flare.txt` to make the tests pass in CI once the integration tests are pushed.

connections/sim/hw_sim.py
```diff
@@ -132,6 +132,27 @@ def read(self):
...
 
+class Clock:
+    def __init__(self) -> None:
+        self._time_us = 0
+
+    def add_time(self, delta_us: float) -> None:
+        """
+        :param delta_us: number of microseconds to add to the clock.
+        """
+        self._time_us += delta_us
```
SilentSys, line 143:
```suggestion
        self._time_us += int(delta_us)
```
Lets always cast the delta to an int to ensure that `self._time_us` doesn't somehow become a float.

SilentSys:
[Method for synchronizing time over SIM](https://trello.com/c/k3XUt7WL/246-method-for-synchronizing-time-over-sim)

connections/sim/hw_sim.py
akoen: Fixed it missed that one.

DSep (commented):
Looks good! Just getting 1 test failing with the change from a few mins ago, but generally SIM seems to run ok. I'm not deeply involved with the clock changes, but I'll try to get a better idea for next week's meeting.

tests/integration_tests/test_sim.py
```diff
@@ -176,6 +176,18 @@ def test_temperature_read(qtbot, main_app):
 
         assert main_app.rocket_data.lastvalue(SubpacketEnum.TEMPERATURE.value) == vals[0]
 
+def test_time_update(qtbot, main_app: MainApp):
```
SilentSys, line 179:
```suggestion
def test_time_update(qtbot, main_app):
```

CI failing because MainApp is not imported. I wouldn't bother importing it just for this type hint.

Also after this is fixed. Is there anything else that needs to be done before we can merge both the PRs?

SilentSys (changes requested):

tests/integration_tests/test_sim.py
akoen: Should be fixed now. Sorry there's been so many simple bugs―I wish I could actually see if my tests pass. Hopefully that's all fixed with your changes.

I've updated the commit hash for this commit on FLARE. I don't think anything else is required before merging.

akoen: Nevermind seems like the Travis build is still failing. Any idea why?

SilentSys:
> Nevermind seems like the Travis build is still failing. Any idea why?

It didn't fail fail. It just timed out because of that python bug I was talking about. My PR will fix that. I'll run all the tests on my machine before merging, but I think its all good.

akoen:
> > Nevermind seems like the Travis build is still failing. Any idea why?
> 
> It didn't fail fail. It just timed out because of that python bug I was talking about. My PR will fix that. I'll run all the tests on my machine before merging, but I think its all good.

Brilliant happy to hear that.

SilentSys:
Passes on my computer.

PR #49, opened 2020-12-28
Added lock on SIM stdin to fix corrupt packet bug

SilentSys:
This solves 1 of the 2 bugs we've been seeing with SIM. Since there wasn't a lock on SIM stdin and there are two possible threads that might try and send packets (SIM thread as a response to time updates, and xbee thread for radio packets), packets would sometimes get mangled together. Usually this would corrupt the SIM packet length field causing the FW to block trying to read more bytes than were sent resulting in the test timeouts we were seeing.


SilentSys:
[SIM integration tests timeout](https://trello.com/c/pJgF3cTs/283-sim-integration-tests-timeout)

akoen (approved):
Works on my end

ADSteele916 (approved):
LGTM

jesseli2002:
LGTM. Was thinking about if it was possible to make the XBee single threaded - you'd need to use a finite state machine (or sequence of them), I think

SilentSys:
> LGTM. Was thinking about if it was possible to make the XBee single threaded - you'd need to use a finite state machine (or sequence of them), I think

Oh I misspoke, theres no xbee thread for sending, only receiving. Its the "Send Thread" and "SIM Thread" that are colliding" They're both necessary.

However, the xbee send callback is not necessary, I will add a commit to remove that. I think it remains from when there was an xbee send thread.

SilentSys:
Just kidding. The send callback is necessary to send the xbee acks from the xbee receive thread.

SilentSys:
But yeah it would be cool if we could get rid of that receive thread somehow. We should keep thinking about good ways to do it and see if something comes up. It's not necessary though.

PR #50, opened 2020-12-31
Git hashes

SilentSys:
- Warn if device registers with wrong required version
- Bundle GS git hash in pyinstaller'd app and display on startup
     - Should hopefully help us get more context if we ever need to analyze a random log file
     - Should help us identify which version we're running when someone is "out in the field" with just the .exe

### DIFF
--- a/build.py
+++ b/build.py
@@ -12,8 +12,8 @@ EXECUTABLE_NAME = 'UBCRGS'
 ENTRY_POINT = 'start.py'
 
 DATA_FILES = [
-    './qt_files/*',
-    './required_flare.txt'
+    'qt_files/*',
+    'required_flare.txt'
 ]
 
 HIDDEN_IMPORTS = [
@@ -57,6 +57,7 @@ PYINSTALLER_SEPARATOR = {
     'darwin': ':'
 }[sys.platform]
 
+GIT_HASH_FILE = os.path.join(LOCAL, '.git_hash')
 
 def _run(executable, args):
     cmd = [executable] + args
@@ -70,6 +71,12 @@ def _is_venv():
     return (hasattr(sys, 'real_prefix') or
             (hasattr(sys, 'base_prefix') and sys.base_prefix != sys.prefix))
 
+def parent_dir(path):
+    path = os.path.dirname(path)
+    if len(path) == 0:
+        return '.'
+    else:
+        return path
 
 def setup_step():
     print("Creating venv...")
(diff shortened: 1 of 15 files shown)


SilentSys:
[Setup version checking with radio config's version id](https://trello.com/c/K2HMJAxP/259-setup-version-checking-with-radio-configs-version-id)

SilentSys:
[Get the current commit hash aka ground station version, and make available to code for future use](https://trello.com/c/QitvHNNa/260-get-the-current-commit-hash-aka-ground-station-version-and-make-available-to-code-for-future-use)

build.py
```diff
@@ -84,11 +92,21 @@ def setup_step():
...
+
+    print(f"Using git hash: {git_hash}")
+    with open(GIT_HASH_FILE, 'w') as f:
+        f.write(git_hash)
+
+    DATA_FILES.append(os.path.relpath(GIT_HASH_FILE, start=LOCAL))
+
     print("Running PyInstaller...")
     hidden_imports = [f"--hidden-import={i}" for i in HIDDEN_IMPORTS]
-    data_files = [f"--add-data={i}{PYINSTALLER_SEPARATOR}{os.path.dirname(i)}" for i in DATA_FILES]
```
jesseli2002, line 89: `os` provides methods to get the absolute directory with no symlinks, from which you can get an absolute path to the parent, but not sure if that's what you were looking for

ADSteele916 (approved):
LGTM

main_window/device_manager.py
```diff
@@ -18,13 +18,16 @@ class DeviceType(Enum):
...
+
         self._lock = RLock()
 
         self._device_type_to_device: Dict[DeviceType, RegisteredDevice] = dict()
         self._full_address_to_device: Dict[FullAddress, RegisteredDevice] = dict()
 
-    def register_device(self, device_type: DeviceType, full_address: FullAddress) -> None:
+    def register_device(self, device_type: DeviceType, device_version: str, full_address: FullAddress) -> None:
         with self._lock:
             if device_type not in self._device_type_to_device.keys() and full_address not in self._full_address_to_device.keys():
```
jesseli2002, line 32: pretty sure you could just do `device_type in self._device_type_to_device` (without `.keys()`), but that can be deferred to a small cleanup PR later - this PR has enough file changes as it is

build.py
SilentSys: We dont need the absolute path, we need the relative path. Also yeah we use the `os` function `os.path.dirname()` for getting a parent dir in the `parent_dir()` function.

But the reason for the `parent_dir()` wrapper around `os.path.dirname(path)` is because `os` returns `''` as the parent dir to `'dir/'`, but pyinstaller expects `'.'`

jesseli2002 (commented):
Can't say I know enough about the architecture to indicate if anythings missing, but nothing low-level seems off. 

One concern I have - what if one of the files being read has a branch name, instead of a git hash? Is support for that method being dropped with this PR?

SilentSys:
> One concern I have - what if one of the files being read has a branch name, instead of a git hash? Is support for that method being dropped with this PR?

Kinda. The GS supports any string for the version, so we stull fully support a branch name. But FLARE only returns a git hash so that's kind of where the limitation comes from.

I don't really see a way around this unless we specify two separate versions: one for CI (where we can use a branch name) and one for version checking. Or maybe we could do something with git tags and semantic versioning.

Lmk if you come up with any good ideas, we can bring it up next meeting. But in general, I don't think we should make this a very complex system with many special cases. Git hashes just work, and when something's wrong, it fails loudly.

main_window/device_manager.py
SilentSys: Good call. I just grepped for other offenders. Looks like ~20 across ~5 files. Will make a PR soon.

PR #54, opened 2021-01-12
Updated UI to make it cleaner and more configurable

SilentSys:
The UI hasn't changed noticeably since the beginning, and it was quite... ugly...

Not it looks a little better and is much more functional IMHO.

Some key changes:

- Every section is now in its own movable and resizable window
     - User's layout preferences are saved and restored on the next session!
- Plot is now in a subwindow
     - Eventually we could add an option for new data graphs to open in new subwindows
- Map resizes to fit display area
- Fixed a few bugs


SilentSys:
[Adjust default sizes and positions of window grid tiles & make map resize with window.](https://trello.com/c/bZxL00Sd/296-adjust-default-sizes-and-positions-of-window-grid-tiles-make-map-resize-with-window)

SilentSys:
[Convert Main window tile grid into MDI?](https://trello.com/c/kxWSvdjo/309-convert-main-window-tile-grid-into-mdi)

SilentSys:
[Label type/unit to string format association](https://trello.com/c/A23UIJ5j/238-label-type-unit-to-string-format-association)

DSep:
Looks awesome, runs well!

DSep (approved):
Looks awesome, runs well!

ADSteele916 (changes requested):
Is there a way to stop "Reset" from sending the window to the top left of my display? Or at least stop it from making the Windows title bar at the top inaccessible? I can't close the application normally after resetting, since I can't get to the Windows exit button without moving the window, which I can't do because the title bar is inaccessible.

ADSteele916:
Also, is it possible to move the controls for the MPL widget to the corner of the widget and not the top right? It doesn't seem right to have it stacked right under the Windows controls in the title bar.

SilentSys:
> Also, is it possible to move the controls for the MPL widget to the corner of the widget and not the top right? It doesn't seem right to have it stacked right under the Windows controls in the title bar.

Unfortunately, unless there's a flag I'm not aware of, I don't think it's possible. It's baked into Qt. A quick google search doesn't return anything. Maybe it could be fixed by applying a theme or something but that seems like overkill.

ADSteele916 (approved):
LGTM. It would be good if we were able to keep the window in the same location upon resetting, but it seems like Qt wouldn't play nicely there.

SilentSys:
> LGTM. It would be good if we were able to keep the window in the same location upon resetting, but it seems like Qt wouldn't play nicely there.

Hm yeah, good point. I think I found a way to make that work with this new commit. Unfortunately it inches down a few pixels every time you reset it, don't think there's a clean way to fix that.

Lmk what you think. And if its better now or before.

ADSteele916:
Definitely better now. This is mergeable, IMO. One thing that I'm just going to note for future reference is that the window will jump to the top left for half a frame or so (sometimes it happens, sometimes it doesn't on my 60 Hz display) before jumping back to the centre. It really doesn't matter, but I just wanted to put that here so we don't misattribute it to any future changes.

jesseli2002:
Also works on my machine.

SilentSys:
> Also works on my machine. Minor complaint - the initial window (selection of connection, rocket, etc.) is smaller and I don't think it's resizable.

Smaller than what? Than before? It shouldn't be... Can you provide screenshots?

What kind of resizing would you want? Same layout just scaling the text?

ADSteele916:
It's shorter than before, but I didn't find it difficult to use or anything.

jesseli2002:
image
vs master:

image

Master also allows it to be resizable

SilentSys:
Oh yeah, that actually looks quite bad/cramped on your system. I'm assuming its Ubuntu. It doesn't look so bad on my Ubuntu VM for some reason. I think I can fix that though. (Commit incoming)

As for the window resizing... I specifically made it non-resizable as it just adds empty space between the controls. Don't see that as being very useful. Lmk if you think there's a good use case for it though.

SilentSys:
It should now no longer squish the controls.

jesseli2002:
yep that worked

PR #55, opened 2021-01-16
Event packet

akoen:
Currently test does not complete.

PR #60, opened 2021-01-31
Update state as per new spec

DSep:
Similar to #55
Needs to have other side done on FLARE repo


main_window/packet_parser.py
```diff
@@ -72,6 +73,18 @@ class SubpacketIds(Enum):
     0x0A: DataEntryValues.EVENT_ABORTED,
     0x0B: DataEntryValues.EVENT_IGNITOR_FIRED,
 }
+STATE_IDS = {
+    0x00: DataEntryValues.STATE_STANDBY,
+    0x01: DataEntryValues.STATE_ARMED,
+    0x02: DataEntryValues.STATE_ASCENT_TO_APOGEE,
+    0x03: DataEntryValues.STATE_POWERED_ASCENT,
+    0x04: DataEntryValues.STATE_MACH_LOCK,
+    0x05: DataEntryValues.STATE_PRESSURE_DELAY,
```
akoen, line 82: Don't forget state PRE_AIR_START_COAST_TIMED

DSep:
Still missing the right required FLARE, yup.

akoen (approved):

connections/debug/debug_connection.py
```diff
@@ -89,7 +90,7 @@ def bulk_sensor_mock_random(self) -> bytearray:
                                          random.uniform(0, 1e6),
                                          random.uniform(49.260565, 49.263859),
                                          random.uniform(-123.250990, -123.246956),
-                                         random.randint(0, 100))
+                                         random.randint(0, 0x09))
```
jesseli2002, line 93: Guessing this is mocking the state, in which case it'd be nice to have this not be a magic number

connections/debug/debug_connection.py
```diff
@@ -126,6 +127,8 @@ def config_mock_set_values(self) -> bytearray:
     def event_mock_set_values(self) -> bytearray:
         return radio_packets.event(self._current_millis(), 0x00)
 
+    def state_mock_set_values(self) -> bytearray:
+        return radio_packets.state(self._current_millis(), 0x05)
```
jesseli2002, line 131: Is there an easy way to grab the name of the state, rather than the magic value?

tests/integration_tests/test_sim.py
```diff
@@ -46,17 +47,17 @@ def set_dummy_sensor_values(sim_app, device_type: DeviceType, sensor_type: Senso
 class TestFlare:
     def test_arming(self, qtbot, sim_app, device_type):
         flush_packets(sim_app, device_type)
-        assert sim_app.rocket_data.last_value_by_device(device_type, DataEntryIds.STATE) == 0
+        assert sim_app.rocket_data.last_value_by_device(device_type, DataEntryIds.STATE) == STATE_IDS[0x00]
```
jesseli2002, line 50: magic number?

jesseli2002 (commented):
Couple of magic numbers which it'd be nice if they weren't there

connections/debug/debug_connection.py
DSep: To go by some enum entry we'd have to either create a local or global reversed dictionary, because the dictionary is number -> DataEntryValue. So this is what you'd prefer? I personally think it's overkill, since it is also the debug connection testing some random state value.
```suggestion
        reversed_dict = {v: k for k, v in STATE_IDS.items()}
        return radio_packets.state(self._current_millis(), reversed_dict[DataEntryValues.STATE_MACH_LOCK])
```

connections/debug/debug_connection.py
DSep: Would you prefer a global containing the range of the STATE_IDS, extracted from the dictionary itself? Not sure what you mean by magic number here though, considering it's a randomized selection of inputs for the bulk debug packet

tests/integration_tests/test_sim.py
DSep: Similar situation to the first comment, also being a testing mechanism too.
OR
For this one I could make it `local variable = 0x00`?

tests/integration_tests/test_debug.py
```diff
@@ -78,17 +81,20 @@ def get_val(val):
...
+    # Special check for state
+    state_val = get_val(DataEntryIds.STATE)
+    assert STATE_IDS[state_input] == state_val
 
     assert LABLES_UPDATED_EVENT.wait(snapshot) >= 1
 
     assert app.AltitudeLabel.text() == '2.00 m'
     assert app.GPSLabel.text() == '9.00000\xb0, 10.00000\xb0'
-    assert app.StateLabel.text() == '11'
+    assert app.StateLabel.text() == 'DataEntryValues.' + STATE_IDS[state_input].name
```
SilentSys, line 97: We can get rid of the `DataEntryValues.` if in profiles/label.py we do `.name` when formatting the state label.

Then I think we can mark this as done as well: https://trello.com/c/A23UIJ5j/238-label-type-unit-to-string-format-association

tests/integration_tests/test_sim.py
SilentSys, line 50: Is there a reason this is better than just using `DataEntryValues.STATE_STANDBY` directly? If not, I think that might be cleaner. Also would address jesse's comment about magic number.

SilentSys (changes requested):
LGTM! Just two small formatting things

tests/integration_tests/test_sim.py
DSep: Great point! Thanks

tests/integration_tests/test_debug.py
DSep: Sounds good. Looks like we should be able to mark it as done, yep.

SilentSys:
[Label type/unit to string format association](https://trello.com/c/A23UIJ5j/238-label-type-unit-to-string-format-association)

SilentSys (approved):
Nice!

SilentSys:
[Update radio State packet as per new spec](https://trello.com/c/qXjDMB2r/317-update-radio-state-packet-as-per-new-spec)

PR #61, opened 2021-02-08
Add time series plotting for labels

ADSteele916:


SilentSys (commented):
Looking great so far. I think its the right way to go.

Made some comments, mostly minor details. I recognize this is just a draft though.

SilentSys:
[Ability to switch from map to time series data plots](https://trello.com/c/3vGFOzpx/233-ability-to-switch-from-map-to-time-series-data-plots)

PR #64, opened 2021-02-22
Rocket data timeset refactor

DSep:
Converted RocketData timeset to use the [SortedContainers](https://github.com/grantjenks/python-sortedcontainers) library.


DSep:
Before merging, consider 2 items:

- Separate the last_value_* and time_series_by_device methods? In the currently implementation, calling last_value_by_device becomes a O(n) operation instead of an O(1) operation, since we have to cycle through all the keys.
- Refactoring the rest of the code using the timeset to be more efficient since there is no need to always get .keys() and sort, the .items() or dictionary iterator itself is already in the sorted format needed.

DSep:
[Refactor rocket data timeset to more efficient ordered dict](https://trello.com/c/xgwzsgdX/321-refactor-rocket-data-timeset-to-more-efficient-ordered-dict)

SilentSys:
Yes and yes! I think those are both great optimizations. Although they add a slight amount of complexity, I think they will be used enough to be worth it.

DSep:
For reference, I may avoid doing the second change, as it may actually make it more confusing at this point. I'll see how it looks first.

main_window/rocket_data.py
```diff
@@ -146,7 +146,7 @@ def time_series_by_device(self, device: DeviceType, data_entry_id: DataEntryIds)
 
             return t, y
 
-    def last_value_and_time(self, device: DeviceType, data_entry_id: DataEntryIds) -> tuple:
+    def last_value_and_time(self, device: DeviceType, data_entry_id: DataEntryIds) -> Optional[tuple]:
         """
         Gets the most recent value and its time for the specified DataEntryIds (enum object)
```
jesseli2002, line 152: Probably document under what situations `None` is returned

main_window/rocket_data.py
```diff
@@ -157,19 +157,22 @@ def last_value_and_time(self, device: DeviceType, data_entry_id: DataEntryIds) -
         :return: Value, Time
         :rtype:
         """
-        ret = self.time_series_by_device(device, data_entry_id)
-
-        if ret is None:
-            return None
+        with self.data_lock:
+            all_times = list(self.timeset.keys())
```
jesseli2002, line 161: `.keys()` supports `reversed()`, so no need to convert into list. Same applies for other list(keys()) call.

main_window/rocket_data.py
```diff
@@ -157,19 +157,22 @@ def last_value_and_time(self, device: DeviceType, data_entry_id: DataEntryIds) -
...
-        t, y = ret
+            full_address = self.device_manager.get_full_address(device)
+            if full_address is None:
+                return None
 
-        if len(t) > 0:
-            return y[-1], t[-1]
+            data_entry_key = DataEntryKey(full_address, data_entry_id)
+            # iterate in reverse time order to get most recent entry
+            for time in reversed(all_times):
```
jesseli2002, line 169: Python dictionaries support iterating over the values only, so this for loop can probably be simplified by choosing a different thing to iterate over

main_window/rocket_data.py
```diff
@@ -157,19 +157,22 @@ def last_value_and_time(self, device: DeviceType, data_entry_id: DataEntryIds) -
...
-
-        if ret is None:
-            return None
+        with self.data_lock:
+            all_times = list(self.timeset.keys())
 
-        t, y = ret
+            full_address = self.device_manager.get_full_address(device)
+            if full_address is None:
+                return None
```
jesseli2002, line 165: Not going to require it, but my understanding is that `None` is acting like an error code, but a more Pythonic approach might be to raise a (custom) DataUnavailable exception.

main_window/rocket_data.py
```diff
@@ -226,16 +229,13 @@ def save(self, csv_path):
...
                     device = self.device_manager.get_device_type(keys[ix].full_address)
-                    if device:
-                        device_name = device.name
-                    else:
-                        device_name = f"{keys[ix].full_address.connection_name}_{keys[ix].full_address.device_address}"
+                    device_name = device.name if device else \
+                        f"{keys[ix].full_address.connection_name}_{keys[ix].full_address.device_address}"
 
                     data[ix, iy] = data_name + '_' + device_name
                 else:
```
jesseli2002, line 241: ok this isn't strictly part of the changed code, but aren't the `else: if: ` blocks below just `elif:` ?

jesseli2002 (dismissed):
Some nitpicky things. 
Any way this class could be unit-tested?

DSep:
Suggestion by @SilentSys: Create another data structure for storing rocket data indexed by DataEntryKey instead of time, and each of their values will contain a time-indexed series of data entry values.

SilentSys:
> Suggestion by @SilentSys: Create another data structure for storing rocket data indexed by DataEntryKey instead of time, and each of their values will contain a time-indexed series of data entry values.

I actually didn't even considering it having a time-indexed series inside. I was just thinking to have it store the latest value only. But your idea is actually great because it could help with some of the other "get" functions.

Actually now that I think about it. Would it be better to refactor the whole data structure as [DataEntryKey][time] instead of the current [time][DataEntryKey]? Is there any use case where the current method is better than the other one? I guess maybe for when we are saving to CSV and group everything together by time...

DSep:
[Cleanup rocket data](https://trello.com/c/JmsHdeeC/264-cleanup-rocket-data)

DSep:
Thanks @jesseli2002 for those comments again. For now, we'll be leaving it as is, since the empty/missing cases are expected and a part of normal operation. There's a trello card keeping track of your comments and your pros and con from slack though, to possibly consider in the future!

main_window/rocket_data.py
```diff
@@ -226,31 +244,30 @@ def save(self, csv_path):
...
-                            value = self.timeset[times[iy - 1]][keys[ix]]
-                            data[ix, iy] = value if not isinstance(value, Enum) else value.name
-                        else:
-                            data[ix, iy] = ""
+                    data[ix, iy] = ""
 
         np.savetxt(csv_path, np.transpose(data), delimiter=',',
                    fmt="%s")  # Can free up the lock while we save since were no longer accessing the original data
 
+    # TODO Missing unit test
```
jesseli2002, line 270: Maybe stick this in a trello ticket

jesseli2002 (approved):

main_window/rocket_data.py
```diff
@@ -26,11 +27,15 @@ def __init__(self, device_manager: DeviceManager) -> None:
...
         self.data_lock = threading.RLock()  # create lock ASAP since self.lock needs to be defined when autosave starts
-        self.timeset: Dict[int, Dict[DataEntryKey, Union[int, float, str]]] = {}
-        self.last_time = 0  # TODO REVIEW/CHANGE THIS, once all subpackets have their own timestamp.
+        # Map: Time -> key-indexed list of data
+        self.timeset: SortedDict[int, Dict[DataEntryKey, Union[int, float, str, Enum]]] = SortedDict()
+        # Map: Key -> time-indexed list of data
+        self.keyset: Dict[DataEntryKey, SortedDict[int, Union[int, float, str, Enum]]] = dict()
+        self.last_time = 0
         self.highest_altitude: Dict[FullAddress, float] = dict()
+        self.existing_entry_keys: Set[DataEntryKey] = set()  # Set of entry keys actually recorded. Used for csv header
```
SilentSys, line 36: Since now we have the `self.keyset` is there any benefit to keeping `self.existing_entry_keys` vs using `self.keyset.keys`?

SilentSys (commented):
All the new changes LGTM! This is great, looking forward to the performance improvements!

Just one question about something that may now be obsolete though.

main_window/rocket_data.py
DSep: That's a good point. I guess I implicitly thought it was important for something but didn't realize it was the same and could likely just be removed.

main_window/rocket_data.py
```diff
@@ -100,8 +108,11 @@ def add_bundle(self, full_address: FullAddress, incoming_data: Dict[DataEntryIds
             # write the data
             for data_id in incoming_data:
                 key = DataEntryKey(full_address, data_id)
-                self.existing_entry_keys.add(key)
+
+                if key not in set(self.keyset.keys()):
```
SilentSys, line 112:
```suggestion
                if key not in self.keyset:
```
I know this works with a normal dict, idk about ordereddict though.

main_window/rocket_data.py
```diff
@@ -222,35 +238,34 @@ def save(self, csv_path):
             # all appearing keys
             keys: List[DataEntryKey] = list(map(
                 lambda data_entry_key: DataEntryKey(data_entry_key.full_address, data_entry_key.data_id),
-                self.existing_entry_keys))
+                set(self.keyset.keys())))
```
SilentSys, line 241:
```suggestion
                self.keyset.keys()))
```
Might not need to be here as well

SilentSys (changes requested):

main_window/rocket_data.py
DSep: Oh yes, sorry for this one .key's is unnecessary.

main_window/rocket_data.py
DSep: I was not able to get .keys() to work without fully converting to set due to its type (set-like).

PR #65, opened 2021-02-28
Add functionality to download map tiles outside of MainWindow

ADSteele916:


SilentSys:
[Implement early download system for mapbox](https://trello.com/c/HVHrijNV/234-implement-early-download-system-for-mapbox)

PR #67, opened 2021-04-08
added handling and tests for pin_mode sent through SIM. Also added a …

jordans6:
…basic template for a generic pin class


connections/sim/hw/pin_sim.py
```diff
@@ -0,0 +1,29 @@
...
+    provides generic pin attributes to be inherited by other pin classes
+    for HW sim
+    """
+
+    INPUT = 0
+    OUTPUT = 1
+
+    def __init__(self):
+        #could be updated later.  For now, sets the pin_mode to be default as an output
+        self.pin_mode = self.OUTPUT
```
SilentSys, line 14: Lets make the default be a 3rd state to indicate that the pin has not been set yet. Then we can also use this in the SIM test to make sure that the FW is actually setting the pins.

connections/sim/sim_connection.py
```diff
@@ -170,6 +171,14 @@ def _handleDigitalPinWrite(self):
         self._hw_sim.digital_write(pin, value)
         LOGGER.info(f"SIM: Pin {pin} set to {value} (device_address={self.device_address})")
 
+    def _handlePinMode(self):
+        length = self._getLength()
+        assert length == 2
+        pin, mode = self.stdout.read(2)
+
+        self._hw_sim.set_pin_mode(pin, mode)
+        LOGGER.info(f"SIM: Pin {pin} set to {mode} (device_address={self.device_address})")
```
SilentSys, line 180: This log is identical to the one in `_handleDigitalPinWrite()` we should somehow differentiate them to prevent confusion in the future.

tests/integration_tests/test_sim.py
```diff
@@ -102,6 +102,23 @@ def test_gps_read(self, qtbot, sim_app, device_type):
...
+            (1, 0),
+            (2, 1),
+        ]
+
+        hw = get_hw_sim(sim_app, device_type)
+        for vals in test_vals:
+            hw.set_pin_mode(vals[0], vals[1])
+            flush_packets(sim_app, device_type)
+
+            assert hw.get_pin_mode(vals[0]) == vals[1]
```
SilentSys, line 119: Currently this is acting as more of a test of hw_sim. However the test_sim.py file is meant more as a test of the FW. Once we have the 3rd pin state representing the default unset state (as per my other comment) we should use this test to check that some pins are being set to something.

connections/sim/hw/hw_sim.py
```diff
@@ -44,6 +61,7 @@ def digital_write(self, pin, val):
             elif pin in self._ignitor_fires and val:
                 self._ignitor_fires[pin].fire()
 
+
     def analog_read(self, pin):
```
SilentSys, line 65: In here we should assert that the pin is set as an input.

connections/sim/hw/hw_sim.py
```diff
@@ -32,6 +35,20 @@ def __init__(
...
+    def get_pin_mode(self, pin):
+        """
+        :param pin: Should be a test pin
+        """
+        with self._lock:
+            val = self._pin_modes[pin]
+            LOGGER.debug(f"Pin mode read from pin={pin} returned value={val}")
+            return val
+
     def digital_write(self, pin, val):
```
SilentSys, line 52: In here we should assert that the pin is set as an output.

SilentSys (changes requested):
Looking good, just a few things to finalize before we merge

connections/sim/hw/hw_sim.py
jordans6: We are now checking to see that the pin is set as an input

bd9f4ee

connections/sim/hw/hw_sim.py
jordans6: We are now checking to see that the pin is set as an output

bd9f4ee

jesseli2002 (changes requested):

connections/sim/hw/hw_sim.py
```diff
@@ -32,26 +35,52 @@ def __init__(
...
+
 
     def analog_read(self, pin):
         """
         :param pin: Should be a read pin. Don't rely on behaviour if the pin isn't a readable pin.
         """
         with self._lock:
             val = 0
-            if pin in self._ignitor_reads:
+            if pin in self._ignitor_reads and self._pin_modes[pin]:
```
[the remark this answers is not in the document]
jordans6: I know there was some discussion about this in the meeting, is it okay if I go ahead and remove these checks for the read?

connections/sim/hw/hw_sim.py
```diff
@@ -32,26 +35,52 @@ def __init__(
...
             LOGGER.debug(f"Digital write to pin={pin} with value value={val}")
-            if pin in self._ignitor_tests:
+            if pin in self._ignitor_tests and (self._pin_modes[pin] == 0):
                 self._ignitor_tests[pin].write(val)
-            elif pin in self._ignitor_fires and val:
+            elif pin in self._ignitor_fires and val and (self._pin_modes[pin] == 0):
                 self._ignitor_fires[pin].fire()
+            else:
+                LOGGER.debug(f"Pin={pin} is not set to correct mode for digital write")
+                # what behaviour should be added if pin is not in correct mode?
```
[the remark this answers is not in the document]
jordans6: This is now an assertion instead of if/else check

dea3eb7

connections/sim/hw/hw_sim.py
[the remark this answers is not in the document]
jordans6: This is now an assertion instead of if/else check

dea3eb7

connections/sim/hw/hw_sim.py
jordans6: This is now an assertion instead of if/else check

dea3eb7

connections/sim/hw/pin_sim.py
jordans6: I've added this now, however the implementation is still just a dictionary member in the hw_sim class.  I can adjust the implementation of PinMode in a later PR

connections/sim/sim_connection.py
jordans6: Adjusted this now as well

tests/integration_tests/test_sim.py
jordans6: Updated to check against pins set in FW

484666c

tests/integration_tests/test_sim.py
```diff
@@ -103,21 +103,18 @@ def test_gps_read(self, qtbot, sim_app, device_type):
...
+        assert hw.get_pin_mode(20) == PinModes.INPUT
+        assert hw.get_pin_mode(21) == PinModes.INPUT
+        assert hw.get_pin_mode(13) == PinModes.INPUT
+        assert hw.get_pin_mode(13) == PinModes.INPUT
+        assert hw.get_pin_mode(16) == PinModes.INPUT
+        assert hw.get_pin_mode(4) == PinModes.INPUT
+        assert hw.get_pin_mode(14) == PinModes.OUTPUT
+        assert hw.get_pin_mode(35) == PinModes.INPUT
+        assert hw.get_pin_mode(17) == PinModes.INPUT
+        assert hw.get_pin_mode(34) == PinModes.OUTPUT
```
jordans6, line 117: In gpio.h, pin 3 - buzzer, is also listed but his doesn't seem to be set anywhere so I haven't included it

connections/sim/hw/hw_sim.py
```diff
@@ -32,23 +42,45 @@ def __init__(
...
                 self._ignitor_fires[pin].fire()
 
+
+
     def analog_read(self, pin):
         """
         :param pin: Should be a read pin. Don't rely on behaviour if the pin isn't a readable pin.
         """
         with self._lock:
+            assert self._pin_modes[pin] == PinModes.OUTPUT
```
jesseli2002, line 83: no need to assert pin mode on analog read

jordans6: Fixed this, but pin mode is still being set to OUTPUT on _handleAnalogRead.  Should I also change this?  It doesn't have any effect other than adding to the pin_modes dictionary in hw_sim

34a53f5

SilentSys:
[Send result of PinMode through S.I.M.](https://trello.com/c/DCS4UdeM/237-send-result-of-pinmode-through-sim)

SilentSys (approved):
LGTM!

PR #68, opened 2021-05-23
Do not use dummy address for XBee

jesseli2002:
Needed for tests to pass on https://github.com/UBC-Rocket/FLARE/pull/165

### DIFF
--- a/connections/sim/hw/xbee_module_sim.py
+++ b/connections/sim/hw/xbee_module_sim.py
@@ -11,11 +11,10 @@ ESCAPE_XOR = 0x20
 XON = 0x11
 XOFF = 0x13
 NEEDS_ESCAPING = (START_DELIMITER, ESCAPE_CHAR, XON, XOFF)
-SOURCE_ADDRESS = bytes.fromhex('0123456789ABCDEF')  # Dummy address
-XBEE_BROADCAST_ADDRESS = bytes.fromhex('000000000000FFFF')
+XBEE_BROADCAST_ADDRESS = bytes.fromhex("000000000000FFFF")
 
-FRAME_PARSED_EVENT = Event('frame_parsed')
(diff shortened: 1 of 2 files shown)


connections/sim/hw/xbee_module_sim.py
```diff
@@ -93,7 +92,8 @@ def send_to_rocket(self, data):
         rx_options = b"\x02"
         self.rocket_callback(
             self._create_frame(
-                FrameType.RX_INDICATOR, SOURCE_ADDRESS + reserved + rx_options + data,
+                FrameType.RX_INDICATOR,
```
SilentSys, line 95: Wait, was this a bug?

SilentSys: If so, im curious what changed that led to you finding it

SilentSys:
Also just as a reminder about our discussion in the meeting: we were talking about possibly including a test that switches GS XBee addresses and sees if the FW responds to the new address.

connections/sim/hw/xbee_module_sim.py
jesseli2002: I believe previously I had the source address hard coded because it wasn't being used; when I removed that because its value now has importance, I found this reference to it

jesseli2002:
> Also just as a reminder about our discussion in the meeting: we were talking about possibly including a test that switches GS XBee addresses and sees if the FW responds to the new address.
> 
> Did we say that needed to be done as a different PR?

It probably should be done as a different PR, to reduce the amount of changes.

jesseli2002:
https://trello.com/c/YX9VMQZd/360-write-integration-test-for-setting-ground-station-address-on-fw-through-config-command for that matter seems like I already made the ticket

SilentSys (approved):

PR #74, opened 2021-10-31
Added integration test for FLARE automatically updating GS address

MarcoYuen17:
Added integration test to check that messages from FLARE are still received after changing the ground station XBee address


MarcoYuen17:
[Integration test for if FLARE automatically updates the ground station address](https://trello.com/c/0QCuOByv/368-integration-test-for-if-flare-automatically-updates-the-ground-station-address)

tests/integration_tests/test_sim.py
```diff
@@ -84,6 +84,24 @@ def test_config_hello(self, qtbot, sim_app, device_type):
...
+
+        assert CONFIG_EVENT.wait(snapshot) == 1
+
+        for connection in sim_app.connections.values():
+            connection._xbee.gs_address = b'\x00\x13\xa2\x00Ag\x8f\xc1'
+
+        sim_app.send_command(device_type.name + ".config")
+        flush_packets(sim_app, device_type)
+
+        assert CONFIG_EVENT.wait(snapshot) == 2
```
SilentSys, line 103: Can you please double check in the code for `.wait` if a new snapshot needs to be taken before calling this? Alternatively, I think you might be able to call it with `wait(snapshot, num_expected=2)`.

I guess `flush_packets` is delaying it long enough that it doesnt even depend on the waiting, so it'll pass anyways.

SilentSys (changes requested):
LGTM, just one small thing

tests/integration_tests/test_sim.py
MarcoYuen17: The code for `.wait` only requires any previously captured snapshot. Resetting `snapshot` here would require the assert to check for a value of 1 both times, but I believe it would make more sense to keep the same snapshot and check for an incremented value after each config event (feel free to correct me if I'm wrong).

Either way, I'll update the second call to `wait()` with the argument `num_expected  = 2`.

SilentSys: Sounds good!

tests/integration_tests/test_sim.py
```diff
@@ -84,6 +84,24 @@ def test_config_hello(self, qtbot, sim_app, device_type):
...
+
+        assert CONFIG_EVENT.wait(snapshot) == 1
+
+        for connection in sim_app.connections.values():
+            connection._xbee.gs_address = b'\x00\x13\xa2\x00Ag\x8f\xc1'
+
+        sim_app.send_command(device_type.name + ".config")
+        flush_packets(sim_app, device_type)
+
+        assert CONFIG_EVENT.wait(snapshot, 2) == 2
```
SilentSys, line 103: These are the parameters: `def wait(self, snapshot, timeout=60, num_expected=1):` So I think that 2 is currently in the timout's place.

```suggestion
        assert CONFIG_EVENT.wait(snapshot, num_expected=2) == 2
```

SilentSys (changes requested):

tests/integration_tests/test_sim.py
MarcoYuen17: Whoops my bad

SilentSys (approved):
LGTM!

PR #75, opened 2021-11-07
Update rocket_data.py

van1011:
Sort CSV headers alphabetically before save

### DIFF
--- a/main_window/rocket_data.py
+++ b/main_window/rocket_data.py
@@ -261,8 +261,11 @@ class RocketData:
                 # Empty entry if this row of data doesn't have a val in this column
                 else:
                     data[ix, iy] = ""
-
-        np.savetxt(csv_path, np.transpose(data), delimiter=',',
+    
+        keysColumn = 0
+        sortedData = data[data[:, keysColumn].argsort()] #sort by keys in column 0
+        
+        np.savetxt(csv_path, np.transpose(sortedData), delimiter=',',
                    fmt="%s")  # Can free up the lock while we save since were no longer accessing the original data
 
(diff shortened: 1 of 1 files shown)


van1011:
[Sort CSV header (keys) alphabetically when saving](https://trello.com/c/5yA67MVE/342-sort-csv-header-keys-alphabetically-when-saving)

main_window/rocket_data.py
```diff
@@ -261,8 +261,11 @@ def save(self, csv_path):
                 # Empty entry if this row of data doesn't have a val in this column
                 else:
                     data[ix, iy] = ""
-
-        np.savetxt(csv_path, np.transpose(data), delimiter=',',
+    
+        keysColumn = 0
+        sortedData = data[data[:, keysColumn].argsort()] #sort by keys in column 0
```
SilentSys, line 266: I looked into it and numpy's "fancy indexing" creates a copy of the matrix. So copying the entire matrix would get slower as more rows get added over time.

I think it would be better if we did the sorting on the `keys` list higher up. Although just now I realized that that list doesn't store strings, so you'll have to also move the logic that converts keys to strings up there.

SilentSys (changes requested):
Kind of a tangent, but as we discussed last meeting, I'm a little anxious about the overall performance of our save function. There is a chance that we will have to run the software for a few hours during a competition and I'd be worried that our data logging effectively stops because it gets too slow. 

If you get a chance, could you please modify the log on line 64 to make it say how long it took to run the `save()` function? That way we can keep an eye on it and see if its going to be an issue.

PR #76, opened 2021-11-07
Add Pylint GitHub Action

joseph784:


.github/workflows/pylint.yml
```diff
@@ -0,0 +1,23 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pylint
+    - name: Analysing the code with pylint
+      run: |
+        pylint `ls -R|grep .py$|xargs`
```
Green-Avocado, line 23:
```suggestion
        pylint $(git ls-files '*.py')
```

fixes

```
************* Module __init__.py
__init__.py:1:0: F0001: No module named __init__.py (fatal)
```

Green-Avocado (changes requested):
See pr https://github.com/joseph784/UBCRocketGroundStation/pull/2

fixes bug https://github.com/actions/starter-workflows/issues/636

Green-Avocado (approved):

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,23 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pylint
+    - name: Analysing the code with pylint
+      run: |
+        pylint $(git ls-files '*.py')
```
SilentSys, line 23: Do you think it would be worth changing this so that this only runs on files that were changed in the branch? Otherwise I think it will always just give us the same warnings about files that aren't really relevant to whichever PR its running on.

SilentSys (commented):
Looks good. So is this going to start working once it gets merged into master or do we need to set up an account or give something access to this repo?

Also wondering if it would be worth consolidating this into CircleCI so that we don't have to maintain two different CI services.

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pylint
```
Green-Avocado, line 20:
```suggestion
        pip install pylint
      - name: Get changed files
        id: changed-files
        uses: tj-actions/changed-files@v11.5
```

Uses https://github.com/tj-actions/changed-files to only get changed files

.github/workflows/pylint.yml
Green-Avocado:
```suggestion
        PYTHONFILES=""
        for file in ${{ steps.changed-files.outputs.all_modified_files }}; do
          if [[ -f $file && $file == *.py ]]
            then
              PYTHONFILES="${PYTHONFILES} $file"
            fi
        done
        echo $PYTHONFILES
        pylint $PYTHONFILES
```

Get output of `changed-files` and filter to match python files

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,23 @@
+name: Pylint
+
+on: [push, pull_request]
```
Green-Avocado, line 3:
```suggestion
on: [pull_request]
```

Green-Avocado: Doesn't need to run on pushes.

Also avoids this note from the `changed-files` action:

> NOTE: :warning:
>
> *   **IMPORTANT:** For `push` events you need to include `fetch-depth: 0` **OR** `fetch-depth: 2` depending on your use case.
> *   When using `persist-credentials: false` with `actions/checkout@v2` you'll need to specify a `token` using the `token` input

Green-Avocado (changes requested):
Avoid running pylint on unchanged files

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,26 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pylint
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v11.5
```
Green-Avocado, line 23:
```suggestion
    - name: Get changed files
      id: changed-files
      uses: tj-actions/changed-files@v11.5
```

indentation is a bit off

Green-Avocado (changes requested):

joseph784:
> Looks good. So is this going to start working once it gets merged into master or do we need to set up an account or give something access to this repo?

No setup is required, it should be good to go once it's in main

> Also wondering if it would be worth consolidating this into CircleCI so that we don't have to maintain two different CI services.

I think this would be easier just doing it this way, but we could consolidate it if you want

PR #78, opened 2021-11-14
Stop black stitched maps from saving to cache

MarcoYuen17:
Ensures that black images generated when a mapbox API key is not present are not cached. This previously resulted in the black stitched maps continuing to show even after an API key was added.


MarcoYuen17:
[Do not save black stitched maps to disk](https://trello.com/c/q1dGNpN0/257-do-not-save-black-stitched-maps-to-disk)

SilentSys (changes requested):
Looks good!

Can you please just rename the variables to something that doesn't contain "pixel"? Its a bit misleading now that you are no longer checking any pixel values.

SilentSys (approved):

PR #80, opened 2021-12-05
Sort csv

van1011:
Sort keys alphabetically


van1011:
[Sort CSV header (keys) alphabetically when saving](https://trello.com/c/5yA67MVE/342-sort-csv-header-keys-alphabetically-when-saving)

SilentSys (changes requested):
Looks good! Just two little things and I think we're ready to merge.

SilentSys (approved):
🚀

PR #83, opened 2022-01-30
Map zoom

van1011:


SilentSys:
Might need to merge master into this branch or rebase your changes onto master and then fix any conflicts with the multiple devices change.

qt_files/comp_app.ui
```diff
@@ -207,6 +207,71 @@
...
+       <height>21</height>
+      </rect>
+     </property>
+     <property name="font">
+      <font>
+       <pointsize>14</pointsize>
+      </font>
+     </property>
+     <property name="text">
+      <string>+</string>
```
MarcoYuen17, line 246: Just curious, is it possible to make the "+" and "-" inc/decrement the zoom?

van1011: Good idea, + and - are buttons now :)

MarcoYuen17 (approved):
Looks great

qt_files/comp_app.ui
```diff
@@ -207,6 +207,67 @@
     </layout>
    </widget>
   </widget>
+  <widget class="QDockWidget" name="dockWidget_4">
```
SilentSys, line 210: If you cause the dockedwidget to be resized, either by resizing the entire window or pulling out the docked widget and resizing it, the slider gets cut out if you go smaller and leaves a lot of empty space if you go bigger.

Would be nice to make it like the buttons where it stretches to fill the empty space. And also like the labels so that there's a limit on how small you can go. I don't remember all the details of how to do this, but I'm sure you can reverse engineer those. I think you might have to add a layout for that to work or something.

qt_files/comp_app.ui
```diff
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>800</width>
-    <height>600</height>
+    <width>886</width>
+    <height>509</height>
```
SilentSys, line 10: Maybe keep the original dimensions or at least round to something clean.

SilentSys (approved):

PR #84, opened 2022-02-05
Add splash screen and file icon

SilentSys:
Added our logo as the splash screen while loading and made the rocket emoji (same as map marker) the .exe's icon.

These changes only apply to the standalone executable builds. See http://confluence.ubcrocket.com/display/AV/build.py+Usage#build.pyUsage-build on how to build.

Note: PyInstaller does not currently support splash screens on MacOS.


MarcoYuen17:
I'm currently seeing the following error when building with either the --only build or --skip setup flags:

error

SilentSys:
> I'm currently seeing the following error when building with either the --only build or --skip setup flags:
> 
> error

As mentioned in our call, that looks like PyInstaller hasn't been updated to the version specified by this PR. The build.py setup step should fix that:
```
python build.py --only setup build
```
or
```
python build.py --skip test
```

MarcoYuen17 (approved):
Looks awesome

PR #85, opened 2022-02-06
Add simulated voltage sensor

MarcoYuen17:
Completed implementation of simulated voltage sensor (and associated test), adapted from #71. This PR attempts to address some of the changes initially requested there.


SilentSys:
Looks like your tests pass, its just a broken dependency that fails later on. #86 should fix it.

tests/integration_tests/test_sim.py
```diff
@@ -134,6 +136,18 @@ def test_pin_mode(self, qtbot, sim_app, device_type):
...
 
+    def test_voltage_reading(self, qtbot, sim_app, device_type):
+        flush_packets(sim_app, device_type)
+
+        snapshot = get_event_stats_snapshot()
+        sim_app.send_command(device_type.name + ".VOLT")
+        assert SINGLE_SENSOR_EVENT.wait(snapshot) == 1
+
+        last_voltage = sim_app.rocket_data.last_value_by_device(device_type, DataEntryIds.VOLTAGE)
+        assert(round(last_voltage, 1) == VoltageSensor.NOMINAL_VOLTAGE)
```
SilentSys, line 147: I think it would be worth after these lines, changing the voltage in VoltageSensor and then repeating these lines again checking for the new value. 

This way we ensure that the firmware is updating the value and isn't just sending the same thing every time.

SilentSys (changes requested):
LGTM! Just one small thing.

MarcoYuen17:
[Implement battery voltage sensor](https://trello.com/c/zQIGazpL/254-implement-battery-voltage-sensor)

connections/sim/hw/sensors/voltage_sensor_sim.py
```diff
@@ -0,0 +1,37 @@
+from connections.sim.hw.sensors.sensor import *
+
+
+class VoltageSensor(Sensor):
+    """
+    Simulates the hardware used for measuring battery voltage.
+    """
+
+    # 920mV is the value that gets converted to 11.6V in battery.cpp in FLARE 21899292dc39015570f795ef9e607081aab57e3e
```
SilentSys, line 9: The value we return in `read()` gets spit out in battery.cpp from the call to `Hal::analogRead(m_batterySensorPin)`. Hal::analogRead is a wrapper around the [arduino analog read](https://www.arduino.cc/reference/en/language/functions/analog-io/analogread/). This returns raw ADC level as an integer between 0 and 2^10-1, not mV. You can see that battery.cpp does the conversion from ADC level to pin voltage first, and then pin voltage to battery voltage.

Some of the variables and comments in this file refer to this value as a "voltage" which is misleading. Others refer to it as a "voltage level" which is a bit better but also ambiguous. I think "ADC level" is clearest, but you can use whatever. Even "voltage level" is okay as long as we have comments that clearly explain what it actually is.

SilentSys (approved):

PR #88, opened 2022-02-13
Fixes for mapping multiple devices task

van1011:


main_window/competition/mapping/mapping_thread.py
```diff
@@ -94,6 +95,12 @@ def setViewedDevice(self, devices) -> None:
         """
         with self.cv:
             self.viewed_devices = devices
+            for device in self.viewed_devices:
+                if self.device_status[device] == False: #no data for this device
```
van1011, line 99: maybe there's a better way to track whether data is available instead of keeping this device_status dictionary. 

If data unavailable, there is a logger warning but no change on the map.

SilentSys: You could totally just call do `self.rocket_data.last_value_by_device(device, DataEntryIds.LATITUDE) is None`. There's no downside to calling that function more times. Looks like it would save a few lines too.

main_window/competition/comp_app.py
```diff
@@ -186,6 +172,22 @@ def setup_subwindow(self):
...
+
+            view_all = QAction("All", self)
+            all_devices = self.rocket_profile.mapping_devices
+            view_all.triggered.connect(lambda i, all_devices = all_devices: self.set_view_device(all_devices))
+            self.map_view_menu.addAction(view_all)
+
+        for device in self.rocket_profile.mapping_devices:
+            view_device = QAction(f'{device}', self)
+            view_device.triggered.connect(lambda i, device=device: self.set_view_device([device]))
+            self.map_view_menu.addAction(view_device)
```
SilentSys, line 189: Looks like integration tests are failing for non-tantalus profiles.
```
>           self.map_view_menu.addAction(view_device)
E           AttributeError: 'CompApp' object has no attribute 'map_view_menu'

main_window/competition/comp_app.py:189: AttributeError
```
I think its because the indentation changed for this for loop.

SilentSys (approved):
🚀

SilentSys:
[Make map support multiple devices at once](https://trello.com/c/FcTn4Syx/378-make-map-support-multiple-devices-at-once)

PR #91, opened 2022-03-14
Log signal strength

van1011:


connections/serial/serial_connection.py
```diff
@@ -21,7 +21,10 @@ def _newData(self, xbee_message):
             message = ConnectionMessage(
                 device_address=str(xbee_message.remote_device.get_64bit_addr()),
                 connection=self,
-                data=xbee_message.data)
+                data=xbee_message.data,
+                signal_strength = int(self.device.get_parameter("DB"),16))
```
van1011, line 25: Need to test if this receives signal strength correctly.

PR #96, opened 2022-05-02
Time series plot

van1011:
Open new window through view menu, switch to data plot by clicking data labels. 
Demo video in trello

Note:
- pressure has no data
- there is no function that returns time series data for max altitude, I can add one if needed
- similarly, no function that returns total acceleration, only acceleration_x, y, z. currently displaying acceleration_x


van1011:
[Ability to switch from map to time series data plots](https://trello.com/c/3vGFOzpx/233-ability-to-switch-from-map-to-time-series-data-plots)

van1011:
[Checkbox under data labels to optionally open new graphs in new subwindows](https://trello.com/c/nJdChA0m/314-checkbox-under-data-labels-to-optionally-open-new-graphs-in-new-subwindows)

main_window/main_app.py
```diff
@@ -19,7 +18,7 @@
 class MainApp(QtWidgets.QMainWindow):
     sig_send = pyqtSignal(str)
 
-    def __init__(self, connections: Dict[str, Connection], rocket_profile: RocketProfile) -> None:
+    def __init__(self, connections: Dict[str, Connection], rocket_profile) -> None:
```
MarcoYuen17, line 21: Why was this change made?

profiles/mpl_funcs.py
```diff
@@ -0,0 +1,108 @@
...
+
+from main_window.competition.comp_app import MAP_MARKER
+from main_window.device_manager import DeviceType
+from main_window.data_entry_id import DataEntryIds
+from main_window.main_app import MainApp
+from textwrap import wrap
+
+from main_window.mplwidget import MplWidget
+
+label_to_dataID = {"Altitude":DataEntryIds.CALCULATED_ALTITUDE,
```
MarcoYuen17, line 12: Can we also include labels for stage 2?

profiles/mpl_funcs.py
```diff
@@ -0,0 +1,108 @@
...
+                   "MaxAltitude":None,
+                   "State":DataEntryIds.STATE,
+                   "Stage2State":DataEntryIds.STATE,
+                   "Pressure":DataEntryIds.PRESSURE,
+                   "Acceleration": DataEntryIds.ACCELERATION_X,
```
MarcoYuen17, line 17: Perhaps add acceleration in the Y and Z directions as well

profiles/mpl_funcs.py
```diff
@@ -0,0 +1,108 @@
...
+
+    self.im = None
+
+    plot_widget.canvas.ax.set_axis_on()
+    plot_widget.canvas.ax.set_aspect('auto')
+
+    plot_time_series(self, plot_widget, device, label_name)
+
+
+def plot_time_series(self, plot_widget: MplWidget, device:DeviceType, label_name:str) -> None:
```
MarcoYuen17, line 83: It would be good to label the axes with the data types and units if possible

main_window/main_app.py
van1011: Import loop. MainApp tries to import RocketProfile but RocketProfile imports MainApp.

profiles/mpl_funcs.py
van1011: Screen Shot 2022-05-16 at 11 03 13 PM

What do you think about putting Y,Z acceleration on the same graph?
Alternatively, I can add a net acceleration column in the rocket data and plot that. 

I changed all scatter plot graphs to connected lines. Hopefully easier to read?

profiles/mpl_funcs.py
van1011: Screen Shot 2022-05-18 at 8 28 56 PM
Good idea, added stage 2 :)

The mappings are basically the same for every rocket, so didn't think having one for each profile would help much.

profiles/mpl_funcs.py
MarcoYuen17: Having them all on the same graph is nice, good idea! Can we also have a checkbox list so the user can choose which out of the 3 are plotted at the same time?

profiles/mpl_funcs.py
```diff
@@ -0,0 +1,128 @@
...
+                                DataEntryIds.ACCELERATION_Z]
+               }
+
+label_unit = {"Altitude": "m",
+               "MaxAltitude": "m",
+               "State": "",
+               "Stage2State": "",
+               "Pressure": "" ,
+               "Acceleration": "g",
+               }
```
SilentSys, line 27: You may be able to include this information as properties in the `Label` class. That way these key strings aren't like "magic numbers" which need to be synchronized in multiple places.

profiles/mpl_funcs.py
van1011: Screen Shot 2022-06-01 at 11 27 12 PM

profiles/mpl_funcs.py
```diff
@@ -0,0 +1,116 @@
...
+
+def plot_time_series(self, plot_widget: MplWidget, device:DeviceType, label_name:str) -> None:
+    """
+            Updates the UI when time series data is available for display
+    """
+    plot_widget.canvas.ax.cla()
+    data_entry_id = self.rocket_profile.label_to_dataID[label_name]
+
+    if label_name == "Acceleration":
+        colors = ["Red", "Blue", "Green"]
```
MarcoYuen17, line 87: Looks like this doesn't match the ordering in accelwidget.py's accel_checkboxes, so checking acceleration Y or Z shows the incorrect colours

profiles/rockets/tantalus.py
```diff
@@ -41,16 +42,95 @@ def buttons(self):
             "Ping Stage 2": "TANTALUS_STAGE_2_FLARE.PING",
         }
 
+    @property
+    def label_to_dataID(self):
```
MarcoYuen17, line 46: The non-Tantalus integration tests are failing because this mapping only exists in Tantalus. Could we move the common mappings for both dataIDs and units to a higher level class, then return those along with rocket-specific mappings in these methods?

profiles/rockets/silvertip.py
```diff
@@ -41,12 +42,12 @@ def buttons(self):
     @property
     def labels(self):
         return [
-            Label(DeviceType.SILVERTIP_FLARE, "Altitude", update_altitude),
-            Label(DeviceType.SILVERTIP_FLARE, "MaxAltitude", update_max_altitude, "Max Altitude"),
-            Label(DeviceType.SILVERTIP_FLARE, "GPS", update_gps),
-            Label(DeviceType.SILVERTIP_FLARE, "State", update_state),
-            Label(DeviceType.SILVERTIP_FLARE, "Pressure", update_pressure),
-            Label(DeviceType.SILVERTIP_FLARE, "Acceleration", update_acceleration),
+            Label(DeviceType.SILVERTIP_FLARE, "Altitude", update_altitude, map_fn=receive_time_series),
```
van1011, line 45: map_fn is used for plotting time series and plotting on the map so I added it for silvertip. map_fn = None by default though so it may cause problems when run with other rocket profiles...

MarcoYuen17 (approved):
Note (as discussed): Getting this working in SIM will be a separate task - https://trello.com/c/JtaJmNDE/420-fix-time-series-plotting-in-sim

PR #99, opened 2022-10-23
Workaround for xbee install error

EdwardJXLi:
To fix the error **Could not find a version that satisfies the requirement digi-xbee==1.1.1** for newer versions of pip, we can pull directly from the github to satisfy the requirement.


MarcoYuen17 (approved):
Apologies - forgot to approve this

PR #100, opened 2022-10-28
Allow CI to use Python 3.10

MarcoYuen17:
Pulls new python versions so that 3.10 can be installed for CI

Looks like mapbox tests are now failing, which is consistent with what we saw using 3.10 locally in the past week


EdwardJXLi (approved):
LGTM!

PR #103, opened 2022-11-25
Create pylint.yml

ruwayd99:
Github action that adds a linting check with each commit.


DSep:
Also check out #76

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,34 @@
...
+      uses: tj-actions/changed-files@v11.5
+    - name: Analysing the code with pylint
+      run: |
+        PYTHONFILES=""
+        for file in ${{ steps.changed-files.outputs.all_modified_files }}; do
+          if [[ -f $file && $file == *.py ]]
+            then
+              PYTHONFILES="${PYTHONFILES} $file"
+              echo $PYTHONFILES
+              pylint $PYTHONFILES
```
MarcoYuen17, line 32: Does this run pylint on the current file as well as previously seen files? I think it'd make more sense to just do `pylint $file`

MarcoYuen17: In this case you wouldn't need PYTHONFILES either, and you can change the echo to just echo the current file as well

ruwayd99: Thanks for letting me know! I'll change that then

MarcoYuen17 (approved):
Looks good to me. Let's merge this and try pushing problematic code to a new branch to see the output sometime

PR #107, opened 2023-09-21
Fix CI package installation failures for APRS integration

MarcoYuen17:
- Added custom versions of aprs & kiss packages
  - Ensure aprs package reads its readme.rst using utf8 encoding
  - Remove aprs requirement from kiss setup
    - Avoids fetching package online without the fix above


ruwayd99:
@MarcoYuen17 Please look into it ty

MarcoYuen17:
> @MarcoYuen17 Please look into it ty

@ruwayd99 could you add a constant somewhere with the value 'nd'? that way if it changes, we'd only need to change it in one spot

